PR #4, opened 2019-06-28
feat(bucket): Add support for bulk delete on bucket list page and set min-height of text-editor globally

tthiscafer:


stacks/spica/src/bucket/pages/index/index.component.scss
```diff
@@ -3,6 +3,9 @@
     min-height: 250px;
     padding: 0 30px;
     align-items: flex-start;
+    .deleteMatTool {
```
thesayyn: can we use dash-case instead of cameCase

PR #5, opened 2019-06-30
docs: draft bazel rules for doc generation

thesayyn:
We will generate our public api documentation with [Dgeni](https://github.com/angular/dgeni). 

This PR contains some draft work for appropriate bazel rules for generating docs with dgeni.


tools/dgeni/templates/controller.template.html
```diff
@@ -0,0 +1,90 @@
+{% import "route.macro.html" as routeMacros -%} {% extends 'base.template.html' -%} {% block body %}
+<style>
```
tthiscafer: should not use internal css

docs/site/src/app/components/toc/toc.component.scss
```diff
@@ -0,0 +1,6 @@
+:host {
+  strong {
+    //font-size: 12px;
```
tthiscafer: meaningless explain line

PR #8, opened 2019-07-01
fix(bucket): Exporting with storage and Importing with _id property added

TolgaYigit:


stacks/api/bucket/bucket.controller.ts
```diff
@@ -72,13 +74,26 @@ export class BucketController {
...
-      });
+      for (let data of dataArray) {
+        data._id = new ObjectId(data._id);
+        await this.bds.insertOne(bucketId, data).then(
+          () => {
+            result.insertedToBucketCollection = result.insertedToBucketCollection + 1;
+          },
+          (err: MongoError) => {
+            if (err.code == 11000) {
+              result.skippedDueToDUplication = result.skippedDueToDUplication + 1;
```
thesayyn: Variable name ```skippedDueToDUplication``` should be ```skippedDueToDuplication```

stacks/api/bucket/bucket.controller.ts
```diff
@@ -103,14 +118,31 @@ export class BucketController {
...
 
-    await Promise.all(
-      bucketIds.map(async id => {
-        await this.bds.find(id).then(data => {
-          const stringifiedData = JSON.stringify(data);
-          archive.append(stringifiedData, {name: `${id}.json`});
-        });
-      })
-    );
+    for (let id of bucketIds) {
```
thesayyn: You can flat this code by getting rid of .then chains.

PR #16, opened 2019-07-05
fix(bucket): export with empty storage field

TolgaYigit:


stacks/api/bucket/bucket.controller.ts
```diff
@@ -157,8 +159,12 @@ export class BucketController {
   private asyncFileRequest(url): any {
     return new Promise(function(resolve, reject) {
       request.defaults({encoding: null}).get(url, (err, res, body) => {
-        let extension = mime.extension(res.headers["content-type"]);
-        resolve({extension: extension, body: body});
+        if (res) {
+          let extension = mime.extension(res.headers["content-type"]);
```
thesayyn: We should prefer `const` instead

PR #55, opened 2019-08-27
feat(common): add support for color input on schema placer

tthiscafer:


thesayyn:
This PR is incomplete. 

Please carefully check if the things below are working on the backend

1- Can backend validate the value of a color input?
2- Does the backend know the format of color type?
3- Can the history keep changes of color input?

PR #58, opened 2019-09-02
feat(function): activate & deactivate support for trigger

tthiscafer:


stacks/api/function/engine/interface.ts
```diff
@@ -41,6 +41,7 @@ export interface Triggers {
 
 export interface Trigger {
   type: string;
+  status: string;
```
thesayyn: Why status is string ?

stacks/spica/src/function/interface.ts
```diff
@@ -69,5 +70,5 @@ export interface Trigger<T = any> {
 }
 
 export function emptySubscription(): Subscription {
-  return {trigger: {type: undefined, options: {}}, url: undefined};
+  return {trigger: {type: undefined, status: true, options: {}}, url: undefined};
```
thesayyn: Do we need status for a subscription ? If we don't please make status property optional or create another interface for it.

thesayyn:
@tthiscafer Tests are failing

stacks/spica/src/function/interface.ts
```diff
@@ -38,14 +38,15 @@ export interface Triggers {
...
+  status?: boolean;
   options: any;
 }
 
 export function emptyFunction(): Function {
   return {
     name: undefined,
     description: undefined,
-    triggers: {default: {type: undefined, options: {}}},
+    triggers: {default: {type: undefined, status: true, options: {}}},
```
thesayyn: Please change `status` with `active`

thesayyn: Also, introduce this property to  function schemes.

PR #62, opened 2019-09-10
fix(common): order of array items and fix object property names

tthiscafer:
This PR contains these changes;

1. add ability for order items of an array
2. prepend root path as name to object properties

### DIFF
--- a/stacks/spica/packages/common/src/input/components/array/array.component.scss
+++ b/stacks/spica/packages/common/src/input/components/array/array.component.scss
@@ -37,6 +37,31 @@
             }
           }
         }
+        .drag-box {
+          cursor: move;
+          .cdk-drag-preview {
+            box-sizing: border-box;
+            border-radius: 4px;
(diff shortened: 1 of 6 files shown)


stacks/spica/packages/common/src/input/components/array/array.component.scss
```diff
@@ -37,6 +37,32 @@
             }
           }
         }
+        .drag-box {
+          cursor: move;
```
thesayyn: Please encapsulate CDK specific classes under this class.

stacks/spica/packages/common/src/input/components/array/array.component.ts
```diff
@@ -76,4 +76,8 @@ export class ArrayComponent implements ControlValueAccessor {
   setDisabledState?(isDisabled: boolean): void {
     this._disabled = isDisabled;
   }
+
+  drop(event: CdkDragDrop<string[]>) {
```
thesayyn: Please disable drag & drop if the input disabled

PR #77, opened 2019-09-18
fix(bucket): default language show on language validation error

tthiscafer:


stacks/spica/src/bucket/pages/add/add.component.html
```diff
@@ -72,7 +72,7 @@ <h6>{{ bucket.description }}</h6>
           >
             <property-language #language slot="after"></property-language>
             <mat-error *ngIf="translateModel?.errors" slot="after">
-              This property's default language is required
+              This property's default language ({{ language.default }}) is required
```
TolgaYigit: Default language ({{ language.default }}) of this property has to be filled.

PR #90, opened 2019-09-27
fix(bucket): return empty response if theres no match

thesayyn:


packages/core/testing/request.ts
```diff
@@ -7,10 +7,14 @@ export class Request {
     return `/tmp/${process.env.BAZEL_TARGET.replace(/\/|:/g, "_")}.sock`;
   }
 
-  options<T>(path: string): Promise<Response<T>> {
+  options<T>(path: string) {
```
TolgaYigit: Why have you removed the returning data type?

thesayyn: It is automatically inferred from return value. No need to declare it explicitly

PR #140, opened 2019-12-27
test(bucket): unit test for bucket watcher

htuna07:


stacks/api/bucket/history/watcher.spec.ts
```diff
@@ -0,0 +1,329 @@
...
+      done();
+    };
+    process.addListener("unhandledRejection", handleError);
+
+    //update spy as it returns same bucket document
+    const updatedDocument = {...firstBucketDocuments[0], title: "new title"};
+    previousDocumentSpy.and.returnValue(Promise.resolve(updatedDocument));
+
+    await module
+      .get(DatabaseService)
```
thesayyn: Please put database service into a variable to reuse. 
It would be easier to read.

htuna07: Please check my last commit.

PR #142, opened 2020-01-06
feat(cli): add login command on spica cli

htuna07:


stacks/cli/src/commands/login.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+    const requestOptions = {
+      method: "GET",
+      uri: `${options["url"]}/passport/identify?password=${password}&identifier=${username}`,
+      json: true
+    };
+
+    try {
+      const token = (await request(requestOptions)).token;
+      fs.writeFile(`${homedir}/.spicarc`, JSON.stringify({token: token}), error => {
```
thesayyn: You should also store the server URL for future uses.

stacks/cli/src/commands/login.ts
```diff
@@ -0,0 +1,64 @@
...
+      return;
+    }
+    const requestOptions = {
+      method: "GET",
+      uri: `${options["url"]}/passport/identify?password=${password}&identifier=${username}`,
+      json: true
+    };
+
+    try {
+      const token = (await request(requestOptions)).token;
```
thesayyn: Can you extract this logic to an authorization service thus we can share it across the codebase?

stacks/cli/src/commands/login.ts
```diff
@@ -0,0 +1,64 @@
...
+        }
+      ]
+    };
+  }
+
+  async run(inputs: CommandLineInputs, options: CommandLineOptions): Promise<void> {
+    const username = inputs[0];
+    const password = inputs[1];
+
+    if (!(username && password)) {
```
thesayyn: You should let the required validator to do this check. 

Please check out the other commands to see a similar check.

stacks/cli/src/commands/login.ts
```diff
@@ -40,25 +40,10 @@ export class LoginCommand extends Command {
...
-      const token = (await request(requestOptions)).token;
-      fs.writeFile(`${homedir}/.spicarc`, JSON.stringify({token: token}), error => {
-        error
-          ? this.namespace.logger.error(error.message)
-          : this.namespace.logger.success("Successfully logged in.");
-      });
-    } catch (error) {
-      this.namespace.logger.error(error.message);
-    }
+    const service = new Service();
```
emre2345: Also would be good updating the name of this service variable.

stacks/cli/src/commands/login.ts
```diff
@@ -7,7 +7,9 @@ import {
   validators
 } from "@ionic/cli-framework";
 import {Command} from "../interface";
-import * as loginService from "../login.service";
+import * as authenticationService from "../authentication.service";
+import * as fileService from "../file.service";
```
emre2345: @htuna07 services are more related with remote services on the web. That's why fileService doesn't seem to be proper naming for me. We might use something of io or something. @thesayyn Is there a common naming convention for io classes in the project? I've seen StdOut in the functions module so we might follow the same naming convention.

stacks/cli/src/authentication.service.ts
```diff
@@ -0,0 +1,7 @@
+import * as httpService from "./http.service";
+
+export function identify(username: string, password: string, serverUrl: string): Promise<any> {
```
emre2345: @htuna07 maybe authenticate or login would be a better naming for this method.

emre2345:
@htuna07 Seems ok to me. We might discuss on names in future when we have some time :)

stacks/cli/src/authentication.service.ts
```diff
@@ -0,0 +1,7 @@
+import * as httpService from "./request";
```
thesayyn: I think we should rename `httpService` to `request`.

stacks/cli/src/commands/login.ts
```diff
@@ -0,0 +1,57 @@
+import {
+  CommandLineOptions,
+  CommandMetadata,
+  CommandMetadataInput,
+  CommandMetadataOption,
+  CommandLineInputs,
+  validators
+} from "@ionic/cli-framework";
+import {Command} from "../interface";
+import * as authenticationService from "../authentication.service";
```
thesayyn: I think we should rename `authenticationService` to `authentication`

PR #144, opened 2020-01-10
feat: implement database queue and enqueuer

thesayyn:
Internally tracked by https://dreamharvesters.atlassian.net/browse/SP-49

### DIFF
--- a/stacks/api/function/enqueuer/src/database.ts
+++ b/stacks/api/function/enqueuer/src/database.ts
@@ -1,0 +1,95 @@
+import {ChangeStream, DatabaseService} from "@spica-server/database";
+import {DatabaseQueue, EventQueue} from "@spica-server/function/queue";
+import {Database, Event} from "@spica-server/function/queue/proto";
+import {Description, Enqueuer} from "./enqueuer";
+
+interface DatabaseOptions {
+  collection: string;
+  type: "INSERT" | "UPDATE" | "REPLACE" | "DELETE";
+}
+
+export class DatabaseEnqueuer extends Enqueuer<DatabaseOptions> {
+  description: Description = {
+    title: "Database",
+    name: "database",
+    icon: "view_agenda",
+    description: "Catch up the events happen in the database."
+  };
+
+  private streams = new Set<ChangeStream>();
+
+  constructor(
+    private queue: EventQueue,
+    private databaseQueue: DatabaseQueue,
+    private db: DatabaseService
+  ) {
+    super();
+  }
+
+  subscribe(target: Event.Target, options: DatabaseOptions): void {
+    const stream = this.db.collection(options.collection).watch(
+      [
+        {
+          $match: {operationType: options.type.toLowerCase()}
+        }
+      ],
+      {fullDocument: "updateLookup"}
+    );
+
+    stream.on("change", rawChange => {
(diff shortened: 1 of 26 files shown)


stacks/api/function/enqueuer/src/database.ts
```diff
@@ -0,0 +1,95 @@
...
+        change.updateDescription.updatedFields = JSON.stringify(
+          rawChange.updateDescription.updatedFields
+        );
+      }
+      const event = new Event.Event();
+      event.target = target;
+      event.type = Event.Type.DATABASE;
+      this.queue.enqueue(event);
+
+      this.databaseQueue.enqueue(event.id, change);
```
emre2345: @thesayyn Is it possible that the invocation works before L.59 because of working on another process?

stacks/api/function/enqueuer/src/database.ts
```diff
@@ -0,0 +1,95 @@
...
+          $match: {operationType: options.type.toLowerCase()}
+        }
+      ],
+      {fullDocument: "updateLookup"}
+    );
+
+    stream.on("change", rawChange => {
+      const change = new Database.Change();
+      change.kind = getChangeKind(rawChange.operationType);
+      change.document = rawChange.fullDocument ? JSON.stringify(rawChange.fullDocument) : undefined;
```
emre2345: @thesayyn Does the "DELETE" option pass the deleted data or does it pass undefined to the function?

stacks/api/function/queue/node/database.ts
```diff
@@ -0,0 +1,61 @@
...
+          reject(new Error(error.details));
+        } else {
+          resolve(event);
+        }
+      });
+    });
+  }
+}
+
+export class Change<T = unknown> {
```
emre2345: @thesayyn What's the point of using unknown for a generic?

stacks/api/function/queue/node/database.ts
```diff
@@ -0,0 +1,61 @@
+import {Database} from "@spica-server/function/queue/proto";
+import * as grpc from "grpc";
+
+export class DatabaseQueue {
+  private client: any;
```
emre2345: @thesayyn I think it would be better defining a more strict type for client.

stacks/api/function/queue/node/database.ts
thesayyn: https://stackoverflow.com/questions/51439843/unknown-vs-any/51439876

thesayyn: Generic parameter is optional here.

stacks/api/function/queue/node/database.ts
thesayyn: Current tools do not provide strict types for the service that is why I left this variable as any. 

Also, type information of the service changes regarding the proto files.

stacks/api/function/enqueuer/src/database.ts
thesayyn: It is null if the event type is DELETE.

stacks/api/function/enqueuer/src/database.ts
thesayyn: the runtime class executes the spawn function asynchronously. It won't be a problem.

PR #149, opened 2020-01-14
fix(bucket): secondary property not found on bucket data index page

tthiscafer:


stacks/spica/src/bucket/pages/index/index.component.ts
```diff
@@ -115,19 +115,23 @@ export class IndexComponent implements OnInit {
 
         setTimeout(() => {
           const propertyNameFirst =
-            this.properties.length > 1 && this.properties[1].name
-              ? this.properties[1].name
-              : undefined;
+            this.properties.length > 0 ? this.properties[1].name : undefined;
```
thesayyn: The condition should be `this.properties.length >= 1`

stacks/spica/src/bucket/pages/index/index.component.ts
```diff
@@ -115,19 +115,23 @@ export class IndexComponent implements OnInit {
...
+          const propertyNameSecond =
+            this.properties.length > 1 ? this.properties[2].name : undefined;
```
thesayyn: The condition should be `this.properties.length >= 2`

stacks/spica/src/bucket/pages/index/index.component.ts
```diff
@@ -115,19 +115,23 @@ export class IndexComponent implements OnInit {
...
           const propertyDataFirst =
-            response.data.length > 0 ? response.data[0][this.properties[1].name] : "";
+            response.data.length > 0 && this.properties.length > 0
+              ? response.data[0][this.properties[1].name]
+              : "";
           const propertyDataSecond =
-            response.data.length > 0 ? response.data[0][this.properties[2].name] : "";
+            response.data.length > 0 && this.properties.length > 1
+              ? response.data[0][this.properties[2].name]
+              : "";
```
thesayyn: These lines should be changed like the other property length checks.

stacks/spica/src/bucket/pages/index/index.component.ts
tthiscafer: I don't need equal operator.

tthiscafer: Only the first feature is assigned here.

stacks/spica/src/bucket/pages/index/index.component.ts
thesayyn: or you can explicitly check if the array has a property at the specified index like `this.properties[2] ? this.properties[2].name : undefined;`

stacks/spica/src/bucket/pages/index/index.component.ts
thesayyn: There may not be more than one property. In that case, this check will fail as expected cause `this.properties.length` will be equal to 1 and that single property will be stored at `this.properties[0]`.

So when you try to read the item at index 1, it will fail

stacks/spica/src/bucket/pages/index/index.component.ts
tthiscafer: I don't need equal operator and condition must be greater than 1

PR #150, opened 2020-01-15
feat(function): implement firehose queue

thesayyn:
https://dreamharvesters.atlassian.net/browse/SP-48


stacks/api/function/enqueuer/test/database.spec.ts
```diff
@@ -22,7 +22,7 @@ describe("DatabaseEnqueuer", () => {
     noopTarget.handler = "default";
 
     eventQueue = jasmine.createSpyObj("eventQueue", ["enqueue"]);
-    databaseQueue = jasmine.createSpyObj("httpQueue", ["enqueue"]);
+    databaseQueue = jasmine.createSpyObj("databaseQueue", ["enqueue"]);
```
emre2345: @thesayyn What's this change?

thesayyn: I've changed name of the spy cause it was written wrong.

emre2345: @thesayyn Ok so it's a fix of a previous implementation and not relevant with the firehose integration, right?

thesayyn: Actually, it is just a name change. It does not affect anything at all. 
It's not relevant to the firehose implementation.

thesayyn: I just wanted to change it to prevent misunderstanding.

emre2345:
@thesayyn I think moving stacks/api/function/queue/node/firehose.ts::createMessage and stacks/api/function/queue/src/firehose.ts::parseMessage into a module would be better as both these functions are relevant with proto messages.

thesayyn:
Node directory should not be coupled with other modules since the node directory only contains the source code that needed to communicate with the server. 

Let me explain with an example;

Currently, we have the node runtime and everything (from the communication/queue perspective) relevant to it is under `queue/node` directory. 
If you introduce a new runtime, let's say dotnetcore, then everything relevant to it which works on the worker side will be under `queue/dotnetcore` directory.

So when you look from this perspective those methods are pretty irrelevant.

Also, Everything in the node directory will be bundled into a single javascript file at the end and if we put a function that works at the API side it will be bundled into the bundle file.

stacks/api/function/enqueuer/src/firehose.ts
```diff
@@ -0,0 +1,142 @@
+import {EventQueue, FirehoseQueue} from "@spica-server/function/queue";
+import {Event, Firehose} from "@spica-server/function/queue/proto";
+import * as url from "url";
+import * as ws from "ws";
+import {Description, Enqueuer} from "./enqueuer";
+import express = require("express");
+
+interface FirehoseOptions {
+  event: "*" | "**" | "connection" | "close" | string;
```
htuna07: @thesayyn What's the meaning of these "*" and "**" event types?

stacks/api/function/runtime/test/node/entrypoint.spec.ts
```diff
@@ -296,11 +297,205 @@ describe("Entrypoint", () => {
...
+      await initializeFn(`export default function({socket, pool}, message) {
+        console.log(arguments);
+        if ( 
+          socket.remoteAddress == "[::1]" && 
+          socket.id == "1" &&
+          pool.size == 21 &&
+          message.name == "test" && 
+          message.data == "test"
+        ) {
+          process.exit(4);
```
htuna07: @thesayyn What's the meaning of '4'?

thesayyn: Um. Actually, it's just an exit code but let me give you the background. 

When a process exits with a "non-zero (a number that is not 0)"  it means the process has failed. 
So you may ask why I didn't exit with code 0 instead of 4?

It's because by default the process will quit with "0 exit code" when the program executes successfully so I wanted to check explicitly that the condition meets and the  process exits with code 4.

stacks/api/function/enqueuer/src/firehose.ts
thesayyn: The wildcard '*' will match with everything. 
The wildcard '**' will only match with 'connection' and 'close' events.

stacks/api/function/runtime/node/entrypoint.ts
```diff
@@ -63,7 +66,35 @@ if (!process.env.EVENT_ID) {
       callArguments[0] = new Change(change);
       break;
     case Event.Type.FIREHOSE:
-      // TODO
+      const firehose = new FirehoseQueue();
```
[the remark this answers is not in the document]
thesayyn: I agree. Also, simply we can use plain functions. However, I have concerns about performance. Since this file is the entry point it should be designed to work as fast as possible. Unlike the other parts, performance matters here. That is why we use terser to eliminate dead code and make other optimizations.

If we use the map to accomplish the same behavior, the javascript engine will take some evaluate the map and the functions inside. Also, I think we'll be fine as long as we keep things where they belong. Maybe in the future, we can introduce separate entry points for every enqueuer/queue type but for now, I think we'll be better if we leave it as is.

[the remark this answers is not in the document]
emre2345: SGTM :+1:

PR #151, opened 2020-01-15
feat(cli): add pull command on spica cli

htuna07:


stacks/cli/src/authentication.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export async function getLoginData(): Promise<LoginData> {
+  const rcFile = JSON.parse((await fs.promises.readFile(utilities.getRcPath())).toString());
+  const loginData: LoginData = {
+    token: rcFile.token,
+    server: rcFile.server
+  };
+  if (!loginData.token || !loginData.server)
+    throw {message: "Token or server information is missing."};
```
thesayyn: Replace with `new Error("Token or server information is missing.")`

stacks/cli/src/commands/pull.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export class PullCommand extends Command {
+  async getMetadata(): Promise<CommandMetadata<CommandMetadataInput, CommandMetadataOption>> {
+    return {
+      name: "pull",
+      summary: "Pull assets from the server.",
+      inputs: [
+        {
+          name: "Output folder",
+          summary: "Name of output folder to store assets.",
```
thesayyn: The article is missing here.
This message should be `Name of the output folder to store assets.`

PR #152, opened 2020-01-15
feat(cli): support for push command

htuna07:


emre2345:
@htuna07 Do we have the same models with interface.ts in the api?

stacks/cli/src/interface.ts
```diff
@@ -18,3 +18,54 @@ export abstract class Command extends BaseCommand<
...
+}
+
+export interface Dependency {
+  [key: string]: string;
+}
+
+export interface Asset {
+  kind: string;
+  metadata: MetaData;
+  spec: Function | any;
```
emre2345: @htuna07 Doesn't this line mean spec is always "any"?

htuna07:
> @htuna07 Do we have the same models with interface.ts in the api?
Actually yes but interface **Function** in api doesn't include **indexPath** and **dependencies** properties.

stacks/cli/src/interface.ts
```diff
@@ -18,3 +18,54 @@ export abstract class Command extends BaseCommand<
...
+export interface MetaData {
+  name: string; //id
+}
+
+export interface LoginData {
+  token: string;
+  server: string;
+}
+
+export interface LoginData {
```
thesayyn: There are two LoginData interfaces in this file. 👎

stacks/cli/src/interface.ts
htuna07: But I didn't know how to define this property before we decide asset types like **Bucket**, **Bucket Data** , **Function**..

htuna07: But yes, it means always "any" like you said. What should we do then ?

emre2345: @htuna07 We should use the types if we've them. You might use Any for now but create a tech story for us to discuss later.

thesayyn:
@htuna07 Can you rebase and resolve conflicts?

PR #153, opened 2020-01-15
feat(bucket): support for realtime changes

thesayyn:


packages/database/realtime/database.service.ts
```diff
@@ -0,0 +1,280 @@
...
+            filter(changes => changes.length > 0)
+          )
+          .subscribe(async events => {
+            let syncedIds: string[];
+
+            if (
+              sortedKeys.length == 1 &&
+              sortedKeys[0] == "_id" &&
+              events.length > 1 &&
+              options.sort._id == -1
```
emre2345: @thesayyn What does this if block mean?

emre2345:
@thesayyn 

Do you think we might add more items into the FilterOptions?

```
export interface FindOptions<T> {
  filter?: FilterQuery<T>;
  sort?: {
    [index: string]: -1 | 1;
  };
  skip?: number;
  limit?: number;
}
```

emre2345:
@thesayyn We should split the `find` function in the packages/database/realtime/database.service.ts

packages/database/realtime/database.service.ts
thesayyn: This check should be removed as well as the other commented lines. 

My initial intention was to make the `_id` sort faster by sorting the _id directly in the process.

thesayyn:
Why?

emre2345:
> Why?

Because it's too long and have too many cascade operations. This makes it hard to understand what's going on in the function.

Also need to go through the logic implemented to find a way to make it more understandable. For instance `getMore` variable doesn't make sense to me.

packages/database/realtime/database.service.ts
emre2345: Ok so let's remove them then.

thesayyn:
In order to do that, we have to simplify the logic first. Also, if you are similar to the syntax, it is not a big deal IMHO.

emre2345:
Generally speaking having a function which has around 270 lines isn't something to do with being accustomed to. This is a rule of thumb having small functions. This function violates single responsibility rule which is one of the most important rules in the software development.

I created a tech story https://dreamharvesters.atlassian.net/browse/SP-140

Let's leave this as it's right now and we look into this later once we complete the urgent feature requests.

emre2345:
@thesayyn Went through all the codes and everything seems fine.

packages/database/realtime/database.service.ts
```diff
@@ -0,0 +1,312 @@
...
+    const changedKeys: string[] = (change.updateDescription.removedFields || []).concat(
+      Object.keys(change.updateDescription.updatedFields || {})
+    );
+    return changedKeys.some(k => sortedKeys.indexOf(k) > -1);
+  }
+
+  return true;
+}
+
+function fetchSortedIdsOfTheCursor<T>(
```
emre2345: @thesayyn Do you think we might need more if checks in this function?

packages/database/realtime/database.service.ts
```diff
@@ -0,0 +1,312 @@
...
+function reverseFilter<T>(filter: FilterQuery<T>) {
+  return Object.keys(filter).map(key => {
+    const value = filter[key];
+    return {
+      [`fullDocument.${key}`]: typeof value == "object" ? {$not: value} : {$ne: value}
+    };
+  });
+}
+
+function getSortedKeys(sort: {[k: string]: -1 | 1}) {
```
emre2345: @thesayyn Can you please explain how keys are sorted with this function?

emre2345:
@thesayyn The last commit makes things much cleaner. Good work.

I kind of understand the approach in the find method. You setup sort, fetchMoreToFillCursor and some other functions and use them later in the find logic. 

I wonder if it would be better moving the function setups to outside of the find function. I felt like all the codes in the find method was the continuous flow of the intended logic.

packages/database/realtime/database.service.ts
```diff
@@ -0,0 +1,312 @@
...
+                fullDocument: "updateLookup"
+              }
+            )
+            .on("change", change => {
+              if (hasChangeAlreadyPresentInCursor(ids, change)) {
+                observer.next({
+                  kind: ChunkKind.Expunge,
+                  document: change.documentKey
+                });
+                ids.delete(change.documentKey._id.toString());
```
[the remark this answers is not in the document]
thesayyn: "toHexString" method is the alias of the "toString" method. Good catch I should make them use the same methods.

packages/database/realtime/database.service.ts
thesayyn: As a matter of fact, this function just returns the sorted keys of the cursor. 
Shall I rename it to `getSortedKeysOfTheCursor`? Would it make it more understandable?

packages/database/realtime/database.service.ts
thesayyn: For now, I don't think that we'll need more "if statements" in this function.

packages/database/realtime/database.service.ts
emre2345: @thesayyn I don't get what "sorted" means. In what manner do we sort the keys and why do we sort them? :)

thesayyn: In order to sort the cursor, the client provides the name(s) of the properties that he/she want to sort. 

A complete sorting condition looks like this
`{test:-1, anotherprop:1}`.

So this method returns only the keys of the sorting as an array.

thesayyn:
@emre2345 This PR was the blocker story of the https://dreamharvesters.atlassian.net/browse/SP-140 tech story. I think we have accomplished what we wanted. 

Could you please move SP-140 to Done if you also think this tech story is completed?

PR #154, opened 2020-01-16
feat(passport): complete api key create, edit, list actions

htuna07:


stacks/spica/src/passport/interfaces/api-key.ts
```diff
@@ -0,0 +1,6 @@
+export interface ApiKey {
+  _id?: string;
```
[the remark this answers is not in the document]
htuna07: We use this optional id to understand the difference between create and update request. If id exist, it means we will send update request to service, if it doesn't exist, it means we will send create request to service.

emre2345:
stacks/spica/src/passport/pages/api-key-add/api-key-add.component.html L.14 -> Doest the input element there allows user to enter an id for the apiKey?

htuna07:
> stacks/spica/src/passport/pages/api-key-add/api-key-add.component.html L.14 -> Doest the input element there allows user to enter an id for the apiKey?

@emre2345 There is a **disabled** property inside of input element.

emre2345:
@htuna07 From the "Clean Code" perspective it's not needed using repetitive words as suffixes. 

In `stacks/spica/src/passport/services/api-key.service.ts` all the function names has "apiKey" at the end. Because of being in the "api-key" service, all these functions suppose to be related with api key operations. That's why we can remove the apiKey from all the functions there.

emre2345:
@htuna07 Also regarding the service implementation, we might have a structure as below:

1. Have an interface which has the functions in the api-key service
2. Implement a mock class, which implements these functions, to use while you're developing before the real service is delivered. 
3. Once you have the real backend service, implement a new class which consumes the real service and change only the injection in the composition root.

emre2345:
@htuna07 Do we use `api-key-index/api-key-index.component.ts ` as the naming for the apiKey listing page?

stacks/spica/src/passport/interfaces/api-key.ts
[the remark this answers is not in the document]
emre2345: @htuna07 Thanks for the information

stacks/spica/src/passport/interfaces/api-key.ts
```diff
@@ -0,0 +1,27 @@
...
+import {Observable} from "rxjs";
+import {IndexResult} from "@spica-server/core";
+import {HttpClient} from "@angular/common/http";
+
+export interface ApiKey {
+  _id?: string;
+  key?: string;
+  name: string;
+  description?: string;
+  policies?: Policy[];
```
thesayyn: Type of this field should be `string[]`

stacks/spica/src/passport/pages/api-key-add/api-key-add.component.html
```diff
@@ -0,0 +1,54 @@
...
+        />
+        <mat-error *ngIf="apiKeyNameModel?.errors?.required">This field is required.</mat-error>
+      </mat-form-field>
+      <mat-form-field>
+        <textarea
+          matInput
+          name="apiKeyDescription"
+          placeholder="Description"
+          [(ngModel)]="apiKey.description"
+          #apiKeyDescriptionModel="ngModel"
```
thesayyn: This reference is unused.

stacks/spica/src/passport/pages/api-key-add/api-key-add.component.scss
```diff
@@ -0,0 +1,21 @@
...
+  mat-form-field {
+    width: 100%;
+    margin-top: 15px;
+  }
+  .resource {
+    padding-left: 30px;
+  }
+  .resource-input {
+    padding-left: 5px;
+  }
```
thesayyn: Change class names to domain-specific ones. Even you may remove them cause I couldn't see where they were used.

stacks/spica/src/passport/pages/api-key-add/api-key-add.component.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+          useValue: {
+            params: of({})
+          }
+        },
+        {
+          provide: Router,
+          useValue: {
+            navigate: () => {}
+          }
+        },
```
thesayyn: Use RouterTestingModule instead.

stacks/spica/src/passport/pages/api-key-add/api-key-add.component.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+  it("should set apiKey as emptyApiKey if this page navigated from add button", () => {
+    expect(component.apiKey).toEqual({
+      name: undefined,
+      active: true
+    });
+  });
+
+  it("should set apiKey which return from service if this page navigated from edit button", async () => {
+    component["activatedRoute"].params = of({id: "1"});
+    component["apiKeyService"].get = () =>
```
thesayyn: You can use spyOn here. 

Example;

```
spyOn(component['activatedRoute'], 'get').and.returnValue();
```

stacks/spica/src/passport/pages/api-key-add/api-key-add.component.ts
```diff
@@ -0,0 +1,47 @@
...
+    this.activatedRoute.params
+      .pipe(
+        filter(params => params.id),
+        switchMap(params => this.apiKeyService.get(params.id)),
+        take(1)
+      )
+      .subscribe(apiKey => {
+        this.apiKey = apiKey;
+      });
+  }
```
thesayyn: This logic will fail silently when the id parameter changes reactively.

You should move this logic into the ngOnInit hook and handle reactively by replacing the take(1) pipe with takeUntil pipe and unsubscribe when the view destroyed.

stacks/spica/src/passport/passport-routing.module.ts
```diff
@@ -50,7 +52,15 @@ const routes: Routes = [
...
-      {path: "settings", component: IdentitySettingsComponent}
+      {path: "settings", component: IdentitySettingsComponent},
+      {
+        path: "api-key",
+        children: [
+          {path: "", component: ApiKeyIndexComponent},
+          {path: "add", component: ApiKeyAddComponent},
+          {path: ":id/edit", component: ApiKeyAddComponent}
+        ]
+      }
```
thesayyn: IdentityGuard, PolicyGuard guards are missing in newly introduced endpoints which will let the user access these pages without required policies.

stacks/spica/src/passport/interfaces/api-key.ts
htuna07: I couldn't understand why we use **string[]** instead of **Policy[]**.

thesayyn: cause we just store ids of policies in there.

htuna07: So we won't need to update api key when something changed on policy. Is that why we store id ?

stacks/spica/src/passport/pages/api-key-add/api-key-add.component.ts
htuna07: Can you give me an example of this **when the id parameter changes reactively** scenario ?

stacks/spica/src/passport/interfaces/api-key.ts
thesayyn: Nope. we just store the id cause the original policy is stored somewhere else. [Relation]

stacks/spica/src/passport/passport-routing.module.ts
htuna07: But in this stage, we don't have any service like **passport:api-index** and action like **index,update,show**.

stacks/spica/src/passport/pages/api-key-add/api-key-add.component.spec.ts
htuna07: I tried these ;
`spyOnProperty(component["activatedRoute"], "params", "get").and.returnValue(of({id: "1"}))`
`spyOn(component["activatedRoute"].params,"get").and.returnValue(of({id: "1"}))`
But the result was same **Property params does not have access type get**.
I didn't want to spend too much time for that.

stacks/spica/src/passport/pages/api-key-add/api-key-add.component.ts
```diff
@@ -0,0 +1,54 @@
...
+    this.activatedRoute.params
+      .pipe(
+        filter(params => params.id),
+        switchMap(params => this.apiKeyService.get(params.id)),
+        takeUntil(this.onDestroy)
+      )
+      .subscribe(apiKey => {
+        this.apiKey = apiKey;
+      });
+  }
```
thesayyn: There is no need to put this logic into a separate function since it is not going to be called from outside.

stacks/api/passport/policies/apikey.fullaccess.json
```diff
@@ -7,7 +7,7 @@
     {
       "effect": "allow",
       "action": "passport:apikey:*",
-      "resource": "passport:apikey/*",
```
thesayyn: Why we needed this change?

stacks/spica/src/passport/services/apikey.service.ts
```diff
@@ -0,0 +1,69 @@
+import {Injectable} from "@angular/core";
```
thesayyn: This is the wrong way to mock services in the hierarchical injector. We don't and can't use interfaces to make abstractions cause they are tripped away after the compilation.

thesayyn: Refer to this article: https://blog.danieleghidoli.it/2016/11/06/testing-angular-component-mock-services/

stacks/api/passport/policies/apikey.fullaccess.json
htuna07: Because it returns response as 'Forbidden action' when sending any HTTP requests with fullaccess api key which has this resource. And other features like strategy and policy has the same resource as only '*'.

stacks/spica/src/passport/services/apikey.service.ts
htuna07: The article says there are two ways to mock service, mocking whole class or mocking only methods that will use on test. If you asked why I mocked the whole class, I was developing with mock class before backend service ready to use so that's why I prefer mocking class instead of mocking methods. On the other hand, I couldn't understand the relation between your first and second comments :)

thesayyn: If you want to mock something, first you create an abstract class to define the methods and structure. After that, you create a mock class that extends that class and you provide this mock class to the hierarchical injector.

thesayyn: Example;

```

abstract class Test {
  abstract do(): string;
}

class MockTest  extends Test{
  do(): string {
   return "MockTest.do";
  }
}


// Module
{
  provide: Test,
  useValue: new MockTest()
}

// Component
class AComp{
  constructor(testService: Test) {
    console.log(testService.do()) // Will print MockTest.do to the console
  }
}
```

stacks/spica/src/passport/pages/apikey-add/apikey-add.component.ts
```diff
@@ -0,0 +1,49 @@
...
+export class ApiKeyAddComponent implements OnInit, OnDestroy {
+  @ViewChild("toolbar", {static: true}) toolbar: TemplateRef<any>;
+  public apiKey: ApiKey = emptyApiKey();
+
+  private onDestroy: Subject<void> = new Subject<void>();
+
+  constructor(
+    private router: Router,
+    private activatedRoute: ActivatedRoute,
+    private apiKeyService: apiKeyService
```
thesayyn: `private apiKeyService: apiKeyService` should be `private apiKeyService: ApiKeyService`.

You need to rename it to ApiKeyService.

stacks/spica/src/passport/pages/apikey-index/apikey-index.component.ts
```diff
@@ -0,0 +1,38 @@
+import {Component, OnInit, TemplateRef, ViewChild} from "@angular/core";
+import {MatPaginator} from "@angular/material";
+import {ApiKey} from "src/passport/interfaces/apikey";
+import {Observable, of, Subject, merge} from "rxjs";
+import {switchMap, map} from "rxjs/operators";
+import {apiKeyService} from "src/passport/services/apikey.service";
+
+@Component({
+  selector: "app-apikey-index",
```
thesayyn: Can you please change the selector to `passport-apikey-index` to preserve the convention.

stacks/spica/src/passport/services/apikey.service.ts
```diff
@@ -0,0 +1,72 @@
...
+import {ApiKey, ApiKeyService} from "../interfaces/apikey";
+import {IndexResult} from "@spica-server/core";
+import {HttpClient} from "@angular/common/http";
+import {of} from "rxjs";
+import {ObjectId} from "bson";
+
+@Injectable({providedIn: "root"})
+export class apiKeyService extends ApiKeyService {
```
thesayyn: Can you rename this to `ApiKeyService`?

PR #155, opened 2020-01-20
feat(passport): introduce apikey endpoints

thesayyn:
https://dreamharvesters.atlassian.net/browse/SP-134


stacks/api/passport/apikey/apikey.controller.ts
```diff
@@ -0,0 +1,75 @@
...
+import {ApiKey} from "./interface";
+
+@Controller("passport/apikey")
+export class ApiKeyController {
+  constructor(private aks: ApiKeyService) {}
+
+  @Get()
+  @UseGuards(AuthGuard(), ActionGuard("passport:apikey:index"))
+  find() {
+    return this.aks.find();
```
htuna07: We need to add filter params to this method like skip and limit for pagination changes.

stacks/api/passport/apikey/apikey.controller.ts
```diff
@@ -0,0 +1,75 @@
...
+  insertOne(@Body(Schema.validate("http://spica.internal/passport/apikey")) apiKey: ApiKey) {
+    delete apiKey._id;
+    apiKey.key = uniqid();
+    return this.aks.insertOne(apiKey).then(r => {
+      apiKey._id = r;
+      return apiKey;
+    });
+  }
+
+  @Post(":id")
```
htuna07: Is there any specific reason to use **Post** for update and insert actions?

stacks/api/passport/apikey/apikey.controller.ts
thesayyn: There was no such requirement as far as I know.

stacks/api/passport/apikey/apikey.controller.ts
thesayyn: This could be helpful: https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html

stacks/api/passport/apikey/interface.ts
```diff
@@ -0,0 +1,10 @@
+import {ObjectId} from "@spica-server/database";
+
+export interface ApiKey {
+  _id?: ObjectId;
+  key?: string;
+  name: string;
+  description?: string;
+  policies: Array<string>;
```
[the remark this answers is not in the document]
thesayyn: What do you mean by ValueObjects?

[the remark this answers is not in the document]
emre2345: https://en.wikipedia.org/wiki/Value_object

Value Objects allows us extendibility for possible future improvements.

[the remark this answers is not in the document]
thesayyn: That would make things hard since we will be storing them in the database by duplicating. Passing things as primitives or plain objects much more maintainable and readable in javascript. We separate data and behavior in most cases.

PR #157, opened 2020-01-24
feat(passport): complete authorization the request with apikey

htuna07:


emre2345:
@htuna07 Can you please share some background information on how the auth strategies work. I might have some further questions based on your explanations

htuna07:
> @htuna07 Can you please share some background information on how the auth strategies work. I might have some further questions based on your explanations

Auth strategies provide you to define how to authorize or authenticate the request sender before the controller starts to run. They allow you to catch HTTP requests, extract data from header, body, query, do your logic and return success or fail to your controller for starting to run or throwing an error.

emre2345:
> > @htuna07 Can you please share some background information on how the auth strategies work. I might have some further questions based on your explanations
> 
> Auth strategies provide you to define how to authorize or authenticate the request sender before the controller starts to run. They allow you to catch HTTP requests, extract data from header, body, query, do your logic and return success or fail to your controller for starting to run or throwing an error.

1. Do we use HttpInterceptors?
2. What's the purpose of passing a text to jwt.strategy.ts or apikey.strategy.ts?
3. Where does JwtStrategy.validate method is called?

PR #159, opened 2020-01-28
feat(passport): complete attach and detach policy to apikey

htuna07:


stacks/api/bucket/history/history.controller.ts
```diff
@@ -1,6 +1,6 @@
 import {Controller, Get, Param, UseGuards} from "@nestjs/common";
 import {ObjectId, OBJECT_ID} from "@spica-server/database";
-import {AuthGuard} from "@spica-server/passport";
+import {AuthGuard} from "../../passport/auth.guard";
```
emre2345: @htuna07 May i ask the reason of this change?

htuna07: I was refactoring all **import {AuthGuard} from "@nestjs/passport"** lines to **import {AuthGuard} from "../auth.guard"** cause of we created our AuthGuard but I realized now, there is no difference between these lines on this file. I can revert it back?

stacks/api/passport/auth.guard.ts
```diff
@@ -15,16 +15,17 @@ export type IAuthGuard = CanActivate & {
...
+function createAuthGuard(type?: string): Type<CanActivate> {
   class MixinAuthGuard<TUser = any> implements CanActivate {
     constructor(@Optional() protected readonly options?: AuthModuleOptions) {
-      this.options = this.options || {};
-      if (!type && !this.options.defaultStrategy) {
-        new Logger("AuthGuard").error(NO_STRATEGY_ERROR);
+      if (Array.isArray(this.options.defaultStrategy)) {
+        throw "Default strategy can not be array.";
+      } else {
+        type = this.options.defaultStrategy;
```
thesayyn: Can you also check if the default strategy has provided? Because it has to be provided in case the client does not send a valid scheme. When the client does send an invalid scheme or does not send it at all, we have to fall back to the default strategy and authorize the user against it.

stacks/api/passport/passport.module.ts
```diff
@@ -29,7 +29,7 @@ class PassportCoreModule {
       module: PassportCoreModule,
       imports: [
         CorePassportModule.register({
-          defaultStrategy: options.defaultStrategy || "jwt",
```
thesayyn: You should provide the default strategy. When the user does not send a valid schema, we have to authorize the user against this strategy.

stacks/api/passport/apikey/apikey.controller.ts
```diff
@@ -70,7 +71,7 @@ export class ApiKeyController {
   }
 
   @Post()
-  @UseGuards(AuthGuard(), ActionGuard("passport:apikey:add"))
+  @UseGuards(AuthGuard(), ActionGuard("passport:apikey:update"))
```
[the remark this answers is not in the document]
htuna07: There is no action as 'update' on apikey.json file. 
Available actions are ;
index, show, update, delete

PR #162, opened 2020-02-06
fix(passport): policy statement action selection

htuna07:
https://dreamharvesters.atlassian.net/browse/SP-100

### DIFF
--- a/stacks/spica/src/passport/pages/policy-add/policy-add.component.ts
+++ b/stacks/spica/src/passport/pages/policy-add/policy-add.component.ts
@@ -132,17 +132,27 @@ export class PolicyAddComponent implements OnInit {
     });
   }
 
+  compareActions(serviceActions: string | string[], selectedActions: string | string[]): boolean {
+    if (!Array.isArray(serviceActions)) {
+      serviceActions = [serviceActions];
+    }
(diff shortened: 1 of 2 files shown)


emre2345:
@htuna07 Can you please give some background information on the issue so the review would be more efficient. For ins. what was causing this bug?

stacks/spica/src/passport/pages/policy-add/policy-add.component.ts
```diff
@@ -132,17 +132,27 @@ export class PolicyAddComponent implements OnInit {
...
+    return serviceActions.every(action => selectedActions.includes(action));
+  }
+
   private tidyUpStatements(policy: Policy): Policy {
     const nStatement = policy.statement.map(statement => {
-      const arrayEquality =
-        Array.isArray(statement.action) &&
-        statement.action.every(value => {
-          return this.services[statement.service].actions.indexOf(value) >= 0;
-        });
```
htuna07: @emre2345 
For example ;
Available actions are 'update' , 'delete' and 'index'. You selected only 'index' and 'delete' action and clicked the save button. This logic should check all available actions are selected or not. But this old logic returns true like all actions selected because available actions include selected actions.

PR #166, opened 2020-02-11
build: put version stamp to build artifacts

thesayyn:


scripts/current_version.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+# Used with Bazel's stamping feature
+echo BUILD_SCM_VERSION ${BUILD_SCM_VERSION_OVERRIDE:-$(git describe --abbrev=7 --tags HEAD)}
```
emre2345: @thesayyn Based on https://git-scm.com/docs/git-describe default value of --abbrev is 7 and it's used to change the default value. This means we might remove it if there's not any special reason for you to put it there.

thesayyn:
I couldn't get it what you mean exactly by changing the reason?

emre2345:
> I couldn't get it what you mean exactly by changing the reason?

Sorry. I wanted to ask the reason of the change :)

thesayyn:
I did those changes because instead forcing the users directly to https protocol, we should let the browser decide whether the connection should be established through http or https.

emre2345:
> It's did those changes because instead forcing the users directly to https protocol, we should let the browser decide whether the connection should be established through http or https.

I see, thanks.

scripts/current_version.sh
thesayyn: It might be different in older versions of git. This is the reason why I have put it there.

emre2345: Makes sense. Thanks.

PR #169, opened 2020-02-12
feat(function): multi trigger enabled on client [SP-161]

htuna07:


stacks/spica/src/function/pages/add/add.component.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+import {FunctionService} from "../../function.service";
+import {EnqueuerPipe} from "../../pipes/enqueuer";
+import {of} from "rxjs";
+import {ActivatedRoute} from "@angular/router";
+import {RouterTestingModule} from "@angular/router/testing";
+import {HttpClientTestingModule} from "@angular/common/http/testing";
+import {FUNCTION_OPTIONS, emptyTrigger} from "../../interface";
+import {SchemeObserver} from "../../../../packages/core/layout/scheme.observer";
+
+describe("Functon Add", () => {
```
emre2345: @htuna07 Functon typo

stacks/spica/src/function/pages/add/add.component.ts
```diff
@@ -178,4 +180,21 @@ export class AddComponent implements OnInit, OnDestroy {
...
+    this.isHandlerDuplicated = false;
+    this.function.triggers.forEach(trigger => {
+      const duplicatedHandler = this.function.triggers.filter(
+        item => item.handler == trigger.handler
+      );
+      if (duplicatedHandler.length > 1) {
+        this.isHandlerDuplicated = true;
+      }
+    });
+  }
```
emre2345: @htuna07 Shared a shorter version of the method. 

```
checkHandlers() {
     this.function.triggers.forEach(trigger => {
       const duplicatedHandler = this.function.triggers.filter(
         item => item.handler == trigger.handler
       );
       this.isHandlerDuplicated = duplicatedHandler.length > 1;
       }
     });
   }
```

Another thing is changing the method so it doesn't modify the state of the class but return the result as a boolean would make it more testable as you wouldn't need setting the `function` variable up of the class.

Basically passing a list as parameter and turning a boolean value would make this function a pure one which is more secure.

htuna07: This logic will set the isHandlerDuplicated true when the last value of an array isn't duplicated, but previous values could be duplicated. 

For example ;
The array that will be checked is `["value1","value1","value2"]`. This logic will set isHandlerDuplicated value false for the first and second items, but the last item will set this value true. 

Also, every change in the handler input fields, this function will be called and set the isHandlerDuplicated value. When the isHandlerDuplicated value changes, the error which is dependent on this boolean value will detect these changes and show an error or not. That's why we store this boolean value as global.

emre2345: @htuna07 I see. Is it possible using https://angular.io/guide/form-validation here?

emre2345: @htuna07 Have you seen the code change proposed in my first comment? It removes some lines without changing the usage of the function.

htuna07: We use the template-driven validation instead of reactive form validation. I tried to use the same validation way. In the template-driven forms, we need to create a validator which is checks this duplication but creating a new validator will cause us to spend more time on this basic validation.

emre2345: I see, thanks for the information.

PR #170, opened 2020-02-13
feat(bucket): handled invalid parameters on bucket-data get requests [SP-141]

htuna07:


packages/core/pipes.spec.ts
```diff
@@ -19,6 +20,12 @@ describe("core pipes", () => {
     expect(result).toEqual([1, 2, 3]);
   });
 
+  it("JSONP pipe should throw BadRequestException when given json is invalid", () => {
+    expect(() => {
+      JSONP.transform('*{"key":"value"}', undefined);
+    }).toThrow(new HttpException("Unexpected token * in JSON at position 0", 400));
```
emre2345: @htuna07 Is it required to have the exact same string in the HttpException or does `toThrow` checks only the type?

htuna07: The given value to `toThrow()` is optional. If value exists, it compares the expected value and actual value strictly. Without any given value, it only checks thrown something or not.

But the `toThrowError()` can check the type of error without comparing expected and actual value.

emre2345: Based on your answer, i understand that it would be enough to use

`toThrow(new HttpException("some string", 400));`

Is that correct?

htuna07: Yes, just like I did. The other option is ;
`toThrowError(HttpException,"some string")`. But there is no chance to compare **statusCode** in this option.

PR #177, opened 2020-02-19
refactor(bucket): introduced services module(SP-190)

emre2345:
After introducing the HookModule, a circular dependency situation came up. BucketModule was dependent on HookModule while being vice versa. 

HookModule requires BucketService which used to be in the BucketModule to get the list of current buckets. This was to allow user selecting particular buckets for triggers on the admin client.

Also turned BucketModule into a dynamic module so we can add a feature flag for hooks in the module.


stacks/api/app.module.ts
```diff
@@ -21,7 +21,7 @@ import {DashboardModule} from "@spica-server/dashboard/dashboard.module";
       defaults: [CREATED_AT, UPDATED_AT]
     }),
     PreferenceModule,
-    BucketModule,
+    BucketModule.create(process.env.ENABLE_BUCKET_HOOKS),
```
thesayyn: Shall we rename it forRoot to follow the convention? 
Also, I think it would be good if we have a BucketOptions which has a property to toggle hooks. What do you think?

We can make it work like below;

```typescript
BucketModule.forRoot({
 hooks: !!process.env.ENABLE_BUCKET_HOOKS
})
```

stacks/api/bucket/BUILD.bazel
```diff
@@ -11,14 +11,15 @@ ts_library(
...
         "bucket-data.controller.ts",
         "bucket-data.service.ts",
         "cache.ts",
         "index.ts",
         "scheduler.ts",
-    ],
+    ] + glob(
+        ["hook/*.ts"],
+        exclude = ["hook/*.spec.ts"],
+    ),
```
thesayyn: Since we have the services separated from the main bucket module, we can make the hook module a separate bazel target and depend on it.

This way we will enforce the separation and more incrementality.

stacks/api/bucket/BUILD.bazel
```diff
@@ -224,3 +199,27 @@ docs(
...
+        "@npm//@types/jasmine",
+    ],
+)
+
+jasmine_node_test(
+    name = "hook-test",
+    deps = [
+        ":hook_test_lib",
+    ],
+)
```
thesayyn: Shall we move this into a new BUILD.bazel file in the hooks module ?

stacks/api/bucket/bucket-data.controller.ts
```diff
@@ -16,9 +16,9 @@ import {Schema} from "@spica-server/core/schema";
 import {FilterQuery, MongoError, ObjectId, OBJECT_ID} from "@spica-server/database";
 import {ActionGuard, AuthGuard} from "@spica-server/passport";
 import * as locale from "locale";
-import {BucketDocument} from "./bucket";
+import {BucketDocument} from "@spica-server/bucket/services/bucket";
```
thesayyn: We should reexport the public references from their individual index.ts file. Thus we can import them like `@spica-server/bucket/services` instead of `@spica-server/bucket/services/bucket`.

stacks/api/bucket/bucket-data.controller.ts
```diff
@@ -16,9 +16,9 @@ import {Schema} from "@spica-server/core/schema";
...
 import {BucketDataService, getBucketDataCollection} from "./bucket-data.service";
-import {BucketService} from "./bucket.service";
+import {BucketService} from "./services/bucket.service";
```
thesayyn: We should import this reference from its barrel file.

stacks/api/bucket/bucket.controller.spec.ts
```diff
@@ -76,7 +77,7 @@ describe("Bucket acceptance", () => {
         CoreTestingModule,
         PassportTestingModule.initialize(),
         DatabaseTestingModule.replicaSet(),
-        BucketModule
+        BucketModule.create(process.env.ENABLE_BUCKET_HOOKS)
```
thesayyn: Test targets should not depend on env variables.  I think we'll be better if we put an immediate value instead of a dynamic environment variable.

stacks/api/bucket/bucket.controller.ts
```diff
@@ -19,9 +19,9 @@ import * as archiver from "archiver";
 import * as fs from "fs";
 import * as mime from "mime-types";
 import * as request from "request";
-import {Bucket, BucketDocument, ImportFile} from "./bucket";
+import {Bucket, BucketDocument, ImportFile} from "@spica-server/bucket/services/bucket";
 import {BucketDataService} from "./bucket-data.service";
-import {BucketService} from "./bucket.service";
+import {BucketService} from "./services/bucket.service";
```
thesayyn: We need to change this import clause to import the reference from `@spica-server/bucket/services`

stacks/api/bucket/bucket.module.ts
```diff
@@ -9,43 +9,53 @@ import {BucketDataService} from "./bucket-data.service";
 import {BucketController} from "./bucket.controller";
 import {BucketSchemaResolver, provideBucketSchemaResolver} from "./bucket.schema.resolver";
 import {CUSTOM_TYPES} from "./bucket.schema.types";
-import {BucketService} from "./bucket.service";
+import {BucketService} from "./services/bucket.service";
 import {BucketCache, provideBucketCache} from "./cache";
 import {DocumentScheduler} from "./scheduler";
+import {HookModule} from "@spica-server/bucket/hook";
+import {ServicesModule} from "@spica-server/bucket/services/bucket.service.module";
```
thesayyn: We need to change this import clause to import the reference from `@spica-server/bucket/services`

stacks/api/bucket/bucket.schema.resolver.spec.ts
```diff
@@ -3,7 +3,7 @@ import {DatabaseTestingModule, DatabaseService, ObjectId} from "@spica-server/da
 import {PreferenceModule, Preference} from "@spica-server/preference";
 import {SchemaModule} from "@spica-server/core/schema";
 import {BucketSchemaResolver} from "./bucket.schema.resolver";
-import {BucketService} from "./bucket.service";
+import {BucketService} from "./services/bucket.service";
```
thesayyn: We need to change this import clause to import the reference from `@spica-server/bucket/services`

stacks/api/bucket/bucket.schema.resolver.ts
```diff
@@ -1,8 +1,8 @@
 import {Injectable} from "@nestjs/common";
 import {Validator} from "@spica-server/core/schema";
 import {ObjectId} from "@spica-server/database";
-import {Bucket} from "./bucket";
-import {BucketService} from "./bucket.service";
+import {BucketDocument, Bucket} from "@spica-server/bucket/services/bucket";
+import {BucketService} from "@spica-server/bucket/services/bucket.service";
```
thesayyn: We need to change these import clauses to import the reference from `@spica-server/bucket/services`

stacks/api/bucket/bucket.controller.spec.ts
emre2345: The intention here was to enable hooks based on the modules readiness. Is there any choices that might give us this result?

emre2345:
@thesayyn Still using env vars in tests. I left them intentional until we discuss my aim there. Please see https://github.com/spica-engine/spica/pull/177#discussion_r381273191

thesayyn:
We have failing tests.

Edit: Now that we have disabled the failing flaky test, we should rebase the target branch and see the test results.

PR #178, opened 2020-02-19
fix(function): serialize number and boolean payloads

thesayyn:


stacks/spica/package.json
```diff
@@ -36,7 +36,7 @@
     "path-to-regexp": "^6.1.0",
     "rxjs": "^6.4.0",
     "rxjs-compat": "^6.4.0",
-    "saturn-datepicker": "^8.0.1",
+    "saturn-datepicker": "8.0.1",
```
thesayyn: I had to pin the dependency to 8.0.1 because of a regression. 

See: https://github.com/SaturnTeam/saturn-datepicker/issues/106

PR #185, opened 2020-02-21
fix(function): do not create logs collection as capped

thesayyn:
Tracked as SP-187 as internally.


emre2345:
@thesayyn How does capped collection causes the mention problem in the issue?

thesayyn:
With never versions of mongodb they have removed the support for deletion in a capped collection.

PR #187, opened 2020-02-24
feat(function): before insert trigger for bucket-data [SP-167]

htuna07:


stacks/api/bucket/hooks/src/enqueuer.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  constructor(
+    private queue: EventQueue,
+    private actionQueue: ActionQueue,
+    private dispatcher: ActionDispatcher
+  ) {
+    super();
+  }
+
+  mapHeaders(data: Object) {
```
thesayyn: You ought to extract this method into an exported function.

stacks/api/bucket/hooks/src/module.ts
```diff
@@ -0,0 +1,67 @@
...
+
+function createSchema(service: BucketService) {
+  return service.find({}).then(buckets => {
+    const scheme: JSONSchema7 = {
+      $id: "http://spica.internal/function/enqueuer/bucket",
+      type: "object",
+      required: ["bucket", "type"],
+      properties: {
+        bucket: {
+          title: "Bucket ID",
```
thesayyn: Shall we remove the `ID` suffix.

stacks/api/bucket/hooks/src/queue.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  result(
+    call: grpc.ServerUnaryCall<Action.Result>,
+    callback: grpc.sendUnaryData<Action.Result.Response>
+  ) {
+    const _callback = this.callbacks.get(call.request.id);
+    if (!this.callbacks.has(call.request.id)) {
+      return callback(new Error(`Queue has no callback with id ${call.request.id}`), null);
+    }
+    this.queue.delete(call.request.id);
```
thesayyn: This line was written by me. 👯‍♂  Shall we remove the callback from the callbacks map instead?

In this case, this line should be something like `this. callbacks.delete(call.request.id);`

stacks/api/bucket/hooks/test/proto/node/action.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+      bucket: "test_bucket",
+      type: 0,
+      headers: [
+        new Action.Header({key: "authorization", value: "APIKEY 123123"}),
+        new Action.Header({key: "strategy-type", value: "APIKEY"})
+      ],
+      document: "document_id"
+    });
+  });
+  it("should create ActionParameters", () => {
```
thesayyn: Unnecessary suffix. Shall we remove `ActionParameters`?

stacks/api/bucket/hooks/src/enqueuer.ts
```diff
@@ -0,0 +1,75 @@
...
+          bucket: options.bucket,
+          type: Action.Action.Type[options.type],
+          document: document
+        }),
+        callback
+      );
+    });
+  }
+
+  unsubscribe(target: Event.Target) {
```
thesayyn: Within this method, we need to `off` (unsubscribe from) the dispatcher to prevent a possible memory leak.

PR #189, opened 2020-02-25
fix(function): pressing enter adds trigger

emre2345:
Internal issue: SP-207

Added type to the button which adds new triggers so it's not triggered with keyup.enter automatically


emre2345:
@thesayyn Pushed without writing the tests as could not manage to send "enter" event to the form. I'll appreciate if you find out the solution :)

Issue: SP-212

PR #191, opened 2020-02-26
feat(function): before index trigger for bucket-data

htuna07:


stacks/api/bucket/bucket-data.controller.ts
```diff
@@ -182,6 +183,18 @@ export class BucketDataController {
...
 
+    if (this.dispatcher && strategyType == "APIKEY") {
+      const hookAggregation = await this.dispatcher.dispatch(
+        {bucket: bucketId.toHexString(), type: "INDEX"},
+        headers
+      );
+      if (Array.isArray(hookAggregation) && hookAggregation.length > 0) {
+        hookAggregation.forEach(item => {
+          aggregation.push(item);
+        });
```
thesayyn: Shall we change this line to `aggregation = aggregation.concat(hookAggregation);`?

PR #196, opened 2020-03-01
feat(cli): Introducing new command to install dependency to all Functions

htuna07:


stacks/cli/src/commands/function/index.ts
```diff
@@ -0,0 +1,15 @@
+import {Namespace, NamespaceMetadata, NamespaceMap} from "@ionic/cli-framework";
+import {DependencyNamespace} from "./dependency";
+
+export class FunctionNamespace extends Namespace {
```
thesayyn: You should extend this namespace from  SpicaNamespace which contains all the things needed for a namespace.
```suggestion
import {SpicaNamespace} from "../../interface";

export class FunctionNamespace extends SpicaNamespace {
```

stacks/cli/src/commands/function/dependency/install.ts
```diff
@@ -0,0 +1,108 @@
...
+          summary: "Install specified npm-package to each function.",
+          type: Boolean
+        }
+      ]
+    };
+  }
+
+  showLoginError() {
+    this.logger.error("You need to login before start this action. To login: ");
+    this.logger.info("spica login <username> <password>");
```
thesayyn: When you extend this namespace from `SpicaNamespace` you will be able to change these lines to 
```suggestion
    this.namespace.logger.error("You need to login before start this action. To login: ");
    this.namespace.logger.info("spica login <username> <password>");
```

PR #208, opened 2020-03-10
refactor: split upsert endpoints 

htuna07:


stacks/api/bucket/bucket-data.service.ts
```diff
@@ -2,11 +2,12 @@ import {Injectable} from "@nestjs/common";
 import {
```
thesayyn: I think we'll be better if we use `BaseCollection` helper for this service. Would it take too much time to do it?

stacks/api/bucket/bucket.controller.spec.ts
```diff
@@ -143,35 +144,9 @@ describe("Bucket acceptance", () => {
...
-        }
-      };
-      const secondBucket = {
-        _id: new ObjectId(),
-        title: "New Bucket",
-        description: "Describe your new bucket",
-        icon: "view_stream",
-        primary: "title",
-        readOnly: false,
+      const firstBucket = (await req.post("/bucket", bucket)).body;
```
thesayyn:
```suggestion
      const {body: firstBucket} = await req.post("/bucket", bucket);
```

thesayyn: IMHO, the suggested change would make it much more readable. What do you think ?

thesayyn: There are too many lines which are like this. We should change them as well if you agree with me too.

stacks/api/bucket/scheduler.ts
```diff
@@ -39,7 +39,8 @@ export class DocumentScheduler {
 
     const publish = () => {
       delete data._schedule;
-      this.bds.replaceOne(bucket, data);
+      data._id && delete data._id;
```
thesayyn:
```suggestion
      delete data._id;
```

stacks/api/bucket/services/bucket.service.ts
```diff
@@ -7,8 +7,8 @@ import {
   FilterQuery,
```
thesayyn: Shall we use `BaseCollection` here as well?  Of course, if it is not going to take too much time.

stacks/api/function/src/function.controller.ts
```diff
@@ -73,19 +73,17 @@ export class FunctionController {
...
   }
 
-  @Patch(":id")
+  @Put(":id")
   @UseGuards(AuthGuard(), ActionGuard("function:update"))
-  updateOne(@Param("id", OBJECT_ID) id: ObjectId, @Body(Schema.validate(generate)) fn: Function) {
-    delete fn._id;
-    return this.fs.findOneAndUpdate({_id: id}, {$set: fn}, {returnOriginal: false});
+  replaceOne(@Param("id", OBJECT_ID) id: ObjectId, @Body(Schema.validate(generate)) fn: Function) {
+    if (fn._id) delete fn._id;
```
thesayyn:
```suggestion
 delete fn._id;
```

stacks/api/passport/apikey/apikey.controller.ts
```diff
@@ -73,26 +73,20 @@ export class ApiKeyController {
...
     @Param("id", OBJECT_ID) id: ObjectId,
     @Body(Schema.validate("http://spica.internal/passport/apikey")) apiKey: ApiKey
   ) {
-    delete apiKey._id;
-    return this.aks.findOneAndUpdate({_id: id}, {$set: apiKey}, {returnOriginal: false}).then(r => {
-      if (!r) {
-        throw new NotFoundException();
-      }
-      return r;
+    if (apiKey._id) delete apiKey._id;
```
thesayyn:
```suggestion
    delete apiKey._id;
```

stacks/api/passport/apikey/apikey.controller.ts
```diff
@@ -73,26 +73,20 @@ export class ApiKeyController {
   @Post()
   @UseGuards(AuthGuard(), ActionGuard("passport:apikey:insert"))
   insertOne(@Body(Schema.validate("http://spica.internal/passport/apikey")) apiKey: ApiKey) {
-    delete apiKey._id;
+    if (apiKey._id) delete apiKey._id;
```
thesayyn:
```suggestion
  delete apiKey._id;
```

stacks/spica/src/bucket/services/bucket.service.ts
```diff
@@ -13,11 +13,16 @@ import * as fromBucket from "../state/bucket.reducer";
...
+  patchHttpOptions: object;
   constructor(
     private http: HttpClient,
     private store: Store<fromBucket.State>,
     private preference: PreferencesService
-  ) {}
+  ) {
+    this.patchHttpOptions = {
+      headers: new HttpHeaders().set("Content-Type", "application/merge-patch+json")
+    };
```
thesayyn: Why we don't move these lines into the patchIndexes method?

stacks/api/bucket/bucket.controller.spec.ts
htuna07: If you do not know the **destructuring assignment** ( I didn't know it before JavaScript ), my style is more readable but it won't take too much time, I can make this change suggestions of course. :)

stacks/api/bucket/bucket-data.service.ts
htuna07: It won't take too much time for this service but the other services like passport, identity has more complicated logics so they might take.

stacks/api/bucket/services/bucket.service.ts
htuna07: Just like bucket-data service, it won't.

stacks/spica/src/bucket/services/bucket.service.ts
htuna07: Drag-drop events call this method for all buckets to patch their indexes, so patchIndexes method shouldn't create a new HttpHeaders object for all calls, isn't it?

stacks/api/bucket/services/bucket.service.ts
htuna07: The hook module uses this service. We need to refactor that module also or we won't prefer BaseCollection for only this service.

stacks/spica/src/bucket/services/bucket.service.ts
thesayyn: As a matter of fact, it should, because we want to not have "states in classes" as much as possible.

thesayyn: Since the client application is distributed out of the box, we don't have to worry about a simple allocation.

htuna07:
@thesayyn Only BucketService extended from BaseCollection, other module services didn't refactored because of time cost.

stacks/api/bucket/scheduler.ts
```diff
@@ -39,7 +39,8 @@ export class DocumentScheduler {
 
     const publish = () => {
       delete data._schedule;
-      this.bds.replaceOne(bucket, data);
+      data._id;
```
thesayyn: Is this line needed ?

stacks/spica/src/bucket/services/bucket.service.ts
```diff
@@ -46,14 +46,34 @@ export class BucketService {
...
 
-  updateMany(buckets: Bucket[]): Observable<Bucket[]> {
-    return this.http.put<Bucket[]>(`api:/bucket`, buckets);
+  patchBucket(id: string, changes: object): Observable<number> {
+    return this.http
+      .patch<number>(`api:/bucket/${id}`, changes, {
+        headers: new HttpHeaders().set("Content-Type", "application/merge-patch+json")
+      })
+      .pipe(
+        tap(modifiedCount => {
```
thesayyn: Why do we get modifiedCount instead of the updated document?

stacks/api/bucket/scheduler.ts
htuna07: It should have been `delete data._id`. Fixed on last commit.

PR #214, opened 2020-03-18
refactor(cli): report progress when installing packages

thesayyn:
Closes SP-200 and SP-231.


thesayyn:
@htuna07 first commit is irrelevant with the commit but I had to make those changes FYI.

stacks/api/function/pkgmanager/node/npm.ts
```diff
@@ -1,29 +1,90 @@
...
+
+        if (chunk.indexOf("timing") != -1) {
+          progress += 1;
+        }
+
+        observer.next(Math.min(Math.round(progressAmountOfAStep * progress), 100));
+
+        stderr += chunk;
+      });
+      proc.on("time", () => console.log("DSADASDASDSAD"));
```
htuna07: Probably you forgot to remove this line.

stacks/api/function/pkgmanager/test/node/npm.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  it("should report progress", done => {
+    const progress = [];
+    npm
+      .install(cwd, "debug")
+      .pipe(distinctUntilChanged())
+      .subscribe({
+        next: p => progress.push(p),
+        complete: () => {
+          // prettier-ignore
```
htuna07: Why did we need to ignore this line? :)

thesayyn: Cause it is much more readable this way.

stacks/api/function/pkgmanager/node/npm.ts
thesayyn: Good catch. Thank you

PR #217, opened 2020-03-19
fix(bucket): return documents whose relation fields are empty

thesayyn:


thesayyn:
@htuna07 First commit is irrelevant to the changes that supposed to have been done in this bugfix.

thesayyn:
Tests are failing. I'll look into them tomorrow.

.bazelrc
```diff
@@ -47,3 +47,7 @@ build --host_force_python=PY2
 run --host_force_python=PY2
 
 common --experimental_allow_incremental_repository_updates
+
+
+# Disable Test Caching
+test --nocache_test_results
```
htuna07: Why do we need to disable test caching?

packages/database/testing/testing.module.ts
```diff
@@ -48,98 +49,75 @@ export class DatabaseTestingModule implements OnModuleDestroy {
...
-      ...config.members[2],
-      hidden: true,
-      priority: 0,
-      slaveDelay: 5,
-      tags: {slaveDelay: "true"}
-    };
-    await dbService.executeDbAdminCommand({replSetReconfig: config});
-    //wait until replicaSet reconfigured
-    await new Promise(resolve => setTimeout(resolve, 10000));
-  }
```
htuna07: History service needs to this delay settings if I'm not wrong.

.bazelrc
thesayyn: Some of the test suits are affecting each other due to resource usages or something else. When a test doesn't run along with the affected test, it passes. That's why we want to run tests every time to ensure tests are passing together.

packages/database/testing/testing.module.ts
thesayyn: It didn't fail when I ran the tests so I decided to delete these lines.

thesayyn:
All checks have passed. God may be with us 🥇

PR #226, opened 2020-03-24
fix(bucket): make history work with translated properties and store dates as date

### LINKED ISSUE #98: Translated properties are not working with history
# No description

thesayyn:
Fixes #98 

Internally tracked as SP-235  and SP-21


stacks/api/bucket/bucket.schema.resolver.spec.ts
```diff
@@ -69,35 +70,29 @@ describe("bucket service", () => {
...
-    await module.close();
-  });
+  afterAll(async () => await module.close());
 
-  it("should resolve bucket", async () => {
-    const response = await schemaResolver.resolve(bucket._id.toHexString());
-    expect(response).toEqual({
+  it("should return the compiled bucket schema", async () => {
+    const jsonSchema = await schemaResolver.resolve(bucket._id.toHexString());
+    console.log(jsonSchema);
```
htuna07: Probably you forgot to delete this console.log

stacks/api/bucket/bucket.schema.types.ts
```diff
@@ -1,45 +0,0 @@
-import {Keyword} from "@spica-server/core/schema";
```
htuna07: Why did we need to delete these custom types?

stacks/api/bucket/bucket.schema.resolver.spec.ts
thesayyn: Yes. but it's okay. No need to remove it because this is a test file.

PR #236, opened 2020-03-30
feat(function): allow users to close database connection manually

thesayyn:


stacks/api/function/packages/database/test/database.e2e.spec.ts
```diff
@@ -22,7 +22,22 @@ describe("Database e2e", () => {
...
+
     expect(db instanceof Db).toBe(true);
+    expect(await connected()).toBe(true);
+  });
+
+  it("should connect to the database", async () => {
+    const db = await database();
+    expect(db instanceof Db).toBe(true);
+    expect(await connected()).toBe(true);
+  });
```
htuna07: Probably you forgot to remove one of these blocks?

stacks/api/function/packages/database/test/database.spec.ts
```diff
@@ -63,7 +74,7 @@ describe("database", () => {
     setEnvironment();
     const db = await database();
     expect(connectSpy).toHaveBeenCalledTimes(1);
-    expect(isConnectedSpy).toHaveBeenCalledTimes(1);
+    expect(isConnectedSpy).toHaveBeenCalled();
```
htuna07: Why did we need to make this change?

thesayyn: isConnectedSpy could be called N times depending on the jasmine's seed number. So we don't care how many times it has been called. We just care if it was called.

PR #237, opened 2020-03-31
feat(passport): check whether the user has the ability to do create action or not

htuna07:


stacks/spica/src/passport/directives/can-interact.directive.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+  });
+
+  it("shouldn't show button if condition is false", () => {
+    expect(setVisibleSpy).toHaveBeenCalledTimes(1);
+    expect(setVisibleSpy).toHaveBeenCalledWith(false);
+    expect(button.style.visibility).toEqual("hidden");
+  });
+
+  it("should change visible property of button when condition changed", async () => {
+    setVisibleSpy.calls.reset();
```
thesayyn, line 56: Why do you reset calls before at the beginning? jasmine does it for you already. As you can see you create a new spy in beforeEach calls.

stacks/spica/src/passport/directives/can-interact.directive.ts
```diff
@@ -9,9 +9,33 @@ export class CanInteractDirective implements OnInit {
...
   }
 }
+
+//This directive for test cases. Use it for modules which uses the actual directive on test files.
+@Directive({selector: "[canInteract]"})
+export class CanInteractDirectiveTest implements OnInit {
+  @HostBinding("style.visibility") _visible = "visible";
+  @Input("canInteract") action: string;
+
+  ngOnInit() {}
```
thesayyn, line 40: You can remove ngOnInit since it is unused.

stacks/spica/src/passport/directives/can-interact.directive.spec.ts
htuna07: All beforeEaches create component and pass the condition value to canInteract directive. It means when the fixture is ready right after component created, canInteract calls this setVisible method inside of ngOnInit method. So, when the **it** blocks ready to start, spy have been called 1 time already. If I don't reset these calls, I have to write `toHaveBeenCalledTimes(2)` cause of the initial call that I explained. If I create this spy inside of these **it** blocks, there will be no chance to test the initial spy call.

PR #240, opened 2020-04-02
fix(passport): production build fails cause of CanInteractDirectiveTest directive

htuna07:


htuna07:
Angular production build waits to determine a module for `CanInteractDirectiveTest` directive. But this directive for only test cases. So, it shouldn't be imported into any module which isn't for test. The solution is defining this directive inside of each spec file which uses it.

PR #241, opened 2020-04-03
feat(webhook): introduce webhook module

thesayyn:


stacks/api/function/webhook/test/schema.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+    await db.collection("testing").insertOne({});
+  });
+
+  it("should report when a collection has been dropped", async done => {
+    const schema = resolver.resolve("http://spica.internal/webhook");
+    const coll = db.collection("test");
+    await coll.insertOne({});
+    schema
+      .pipe(
+        bufferCount(2),
```
htuna07, line 82: Is this necessary in this case ?

thesayyn: Yes, it is. You'd see the reason if you take a look at `stacks/api/function/webhook/test/schema.spec.ts:78`.

stacks/api/function/webhook/src/webhook.controller.ts
```diff
@@ -0,0 +1,82 @@
...
+          data: [{$skip: skip}, {$limit: limit}]
+        }
+      },
+      {
+        $set: {
+          meta: {$ifNull: [{$arrayElemAt: ["$meta", 0]}, {total: 0}]}
+        }
+      }
+    ];
+    return this.webhookService.aggregate(aggregate).next();
```
htuna07, line 52: What's the reason of this `.next()` instead of `.toArray()` ?

thesayyn: Cause we don't expect array of documents.

stacks/api/function/webhook/src/webhook.service.ts
```diff
@@ -0,0 +1,73 @@
...
+
+      const stream = this._coll.watch([], {
+        fullDocument: "updateLookup"
+      });
+      stream.on("change", change => {
+        switch (change.operationType) {
+          case "replace":
+          case "update":
+            emitHook(change.fullDocument as Webhook, ChangeKind.Updated);
+            break;
```
htuna07, line 44: If we use `findOneAndReplace` in the PUT endpoint, this logic should be inside of replace instead of update case, am I wrong?

thesayyn: Nope. We don't care if the operation was an update or replace operation since we do a full clean up after something has changed in the database. So this logic does not and shouldn't rely on the type of the update operation (atomic or non-atomic).

stacks/api/function/webhook/test/invoker.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+        name: "database",
+        active: true,
+        options: {
+          collection: "stream_coll",
+          type: "INSERT"
+        }
+      }
+    });
+    await stream.change.wait();
+    await service.updateOne({_id: hook._id}, {$unset: {"trigger.active": ""}});
```
htuna07, line 95: Also, we should change this line too, if we will change the switch cases that I wrote in my previous commit.

stacks/api/function/webhook/test/webhook.service.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+          },
+          {
+            kind: ChangeKind.Removed,
+            target: hook._id.toHexString()
+          }
+        ]);
+        done();
+      });
+    await stream.wait();
+    await service.updateOne({_id: hook._id}, {$set: {"trigger.active": false}});
```
htuna07, line 171: Same changes should be applied to here too, if we change switch cases.

stacks/api/function/webhook/src/webhook.service.ts
thesayyn: Do you have any suggestions about what should be changed here?  and how will it improve the logic here?

thesayyn: @htuna07 ?

htuna07: I don't have any suggestions about this logic because it's clear enough. I was thinking about the place of this logic should be `replace` but you explained this situation. That's all :)

PR #243, opened 2020-04-06
feat(webhook): implementation of the webhook module to the client

htuna07:


stacks/spica/src/function/function-routing.module.ts
```diff
@@ -35,11 +35,11 @@ const routes: Routes = [
...
   {
-    id: "subscription",
+    id: "webhook",
     category: RouteCategory.Developer,
     icon: "http",
-    path: "/subscription",
-    display: "Subscription",
+    path: "/webhook",
+    display: "Webhook",
     data: {action: "subscription:index"}
```
thesayyn, line 43: You have forgotten to change `subscription:index`

PR #250, opened 2020-04-08
fix(storage): can't upload file to storage on storage page on safari browser

htuna07:


stacks/spica/src/storage/pages/index/index.component.ts
```diff
@@ -48,6 +49,12 @@ export class IndexComponent implements OnInit {
   }
 
   ngOnInit(): void {
+    //NgModelChange doesn't work on Safari browsers
+    let fileElement = this.fileSelector.nativeElement;
+    fileElement.addEventListener("change", () => {
+      this.uploadStorageMany(fileElement.files);
+    });
+
```
thesayyn, line 57: It should be working on every major browser. 
Can you point out me to the issue so I can take look?  Is there any open issue about this?

stacks/spica/src/storage/pages/index/index.component.html
```diff
@@ -33,14 +33,7 @@ <h4 class="mat-h4">
...
   >
-    <input
-      type="file"
-      #fileSelector
-      multiple
-      #fileModel="ngModel"
-      [ngModel]="undefined"
-      (ngModelChange)="uploadStorageMany(fileSelector.files)"
-    />
+    <input type="file" #fileSelector multiple #fileModel="ngModel" [ngModel]="undefined" />
```
thesayyn, line 36:
```suggestion
    <input type="file" #fileSelector multiple #fileModel="ngModel" ngModel/>
```

stacks/spica/src/storage/pages/index/index.component.ts
htuna07: Actually it should be `//event binding on file input doesn't work on Safari browsers`.
Here is the example;
https://stackoverflow.com/questions/58341044/input-event-on-file-type-not-firing-in-safari-browser

thesayyn: ngModelChange works in Safari. See: https://stackblitz.com/edit/angular-mfyfa9?file=src%2Findex.html

PR #251, opened 2020-04-08
refactor(function): handle overflow in auto id generation

htuna07:


stacks/api/function/queue/src/event.ts
```diff
@@ -41,7 +41,7 @@ export class EventQueue {
 
   enqueue(event: Event.Event) {
     // TODO: Handle overflow
```
thesayyn, line 43: Could you remove this comment line?

PR #255, opened 2020-04-09
feat(bucket): add ability to filter results on relation picker dialog

htuna07:


thesayyn:
Could you provide some test cases for these changes?

PR #256, opened 2020-04-09
feat(passport): allow user to update own identity without identity access

htuna07:


stacks/api/passport/auth.guard.ts
```diff
@@ -44,16 +44,33 @@ function createAuthGuard(type?: string): Type<CanActivate> {
...
+        user = this.attachIdentityAccess(user);
+
       request[options.property || defaultOptions.property] = user;
 
       request.headers["strategy-type"] = strategyType;
 
       return true;
     }
 
+    isUpdateOwnIdentity(method: string, url: string, user: any) {
```
thesayyn, line 61: We need a more generic and maintainable solution here. ActionGuard can accept a factory function that mutates the policies array on demand.

thesayyn:
Looks good to me. Could you also provide some tests for this?

PR #259, opened 2020-04-15
feat(bucket): displaying relation primary fields on bucket-data index page

htuna07:


stacks/spica/src/bucket/pages/index/index.component.html
```diff
@@ -196,7 +196,14 @@ <h6>{{ schema.description }}</h6>
           <mat-header-cell *matHeaderCellDef mat-sort-header>
             {{ property.value.title }}
           </mat-header-cell>
-          <mat-cell *matCellDef="let row"> {{ row[property.key] }}</mat-cell>
+
+          <mat-cell *matCellDef="let row">
+            {{
+              row[property.key] && row[property.key][property.value.primary]
+                ? row[property.key][property.value.primary]
+                : row[property.key]
```
thesayyn, line 204: Can you test this logic with a related bucket whose primary field contains falsy values?

thesayyn: This logic is going to fail if the related row contains a falsy value which is not what we want. Am i right?

PR #265, opened 2020-04-20
feat(bucket): support for advanced filters

thesayyn:


stacks/spica/src/bucket/components/filter/filter.component.spec.ts
```diff
@@ -29,7 +32,7 @@ class StringPlacer implements ControlValueAccessor {
   registerOnTouched = jasmine.createSpy("registerOnTouched");
 }
 
-describe("FilterComponent", () => {
+fdescribe("FilterComponent", () => {
```
htuna07, line 35: Forgotten `f` character.

PR #269, opened 2020-04-20
fix(bucket): filtering on relation field doesn't work

htuna07:


stacks/spica/src/bucket/components/filter/filter.component.ts
```diff
@@ -42,8 +42,12 @@ export class FilterComponent implements OnChanges {
...
 
   apply() {
-    const origin = this.resolver.getOriginByType(this.properties[this.property].type);
-    this.filter = {[this.property]: origin == "string" ? {$regex: this.value} : this.value};
+    if (this.properties[this.property].type == "relation") {
+      this.filter = {[`${this.property}._id`]: this.value};
+    } else {
+      const origin = this.resolver.getOriginByType(this.properties[this.property].type);
+      this.filter = {[this.property]: origin == "string" ? {$regex: this.value} : this.value};
+    }
```
thesayyn, line 50: Could you provide some tests for these lines?

PR #271, opened 2020-04-22
feat(bucket): adding option to disable/enable history for each bucket on bucket create page

htuna07:


stacks/api/bucket/history/history.service.ts
```diff
@@ -19,6 +19,10 @@ export class HistoryService {
     this.collection = this.db.collection<History>("history");
   }
 
+  getSchema(bucketId: ObjectId) {
+    return this.db.collection<any>("buckets").findOne({_id: bucketId});
+  }
```
thesayyn, line 24: Could you use BucketService directly?

stacks/api/bucket/history/watcher.spec.ts
```diff
@@ -300,4 +310,27 @@ describe("Watcher", () => {
...
+      ...firstBucketDocuments[0],
+      title: "updated title",
+      description: "updated description"
+    };
+
+    await databaseService
+      .collection(`bucket_${buckets[0]._id}`)
+      .replaceOne({_id: updatedDocument._id}, updatedDocument);
+
+    await new Promise(resolve => setTimeout(resolve, 100));
```
thesayyn, line 329: This may cause this test to be flaky. 

Could you please use stream helpers from the `@spica-server/database/testing` module?

stacks/api/bucket/history/history.service.ts
htuna07: Why? Any methods in this service don't use another service directly and there is the comment line like `// We can not use BucketDataService as a direct dependency`

PR #273, opened 2020-04-22
fix(functions): functions doesn't use the pooled processes

emre2345:
https://dreamharvesters.atlassian.net/browse/SP-301


thesayyn:
We should provide some tests for this.  I'll provide them.

PR #275, opened 2020-04-24
refactor(bucket): unset removed field from bucket-data when bucket schema updated

htuna07:


stacks/api/bucket/bucket.controller.ts
```diff
@@ -262,4 +266,73 @@ export class BucketController {
...
+    );
+    if (removedKeys.length < 1) return;
+
+    let unsetFields = removedKeys.reduce((pre, acc: any, index, array) => {
+      acc = {...pre, [array[index]]: ""};
+      return acc;
+    }, {});
+
+    await bucketDataService.updateMany(previousSchema._id, {}, {$unset: unsetFields});
+  }
```
thesayyn, line 289: Can you move these helpers into a new module?

stacks/api/bucket/bucket.controller.ts
```diff
@@ -262,4 +266,73 @@ export class BucketController {
...
+    } else if (this.isObject(previousSchema) && this.isObject(currentSchema)) {
+      this.findRemovedKeys(previousSchema.properties, currentSchema.properties, removedKeys, path);
+    }
+  }
+
+  isObject(schema: any) {
+    return schema &&
+      schema.type == "object" &&
+      schema.properties &&
+      Object.keys(schema.properties).length > 0
```
thesayyn, line 330: Why do we check for the presence of properties?

stacks/api/bucket/bucket.controller.ts
```diff
@@ -262,4 +266,73 @@ export class BucketController {
...
+      ? true
+      : false;
+  }
+
+  isArray(schema: any) {
+    return schema && schema.type == "array" && schema.items ? true : false;
```
thesayyn, line 336: Why do we check for the presence of items?

stacks/api/bucket/bucket.controller.ts
htuna07: Actually it was for `Object.keys(schema.properties).length` might throw an error like `Cannot convert undefined or null to object` when the properties undefined. Is it possible to be undefined?

thesayyn: I'm not sure whether it can be or not.  I don't get that why we even check for properties? 
If its type is 'object' then it should be treated as an object right?

stacks/api/bucket/bucket.controller.ts
htuna07: The same check like object. If it's not possible to be undefined, I can remove it.

stacks/api/bucket/bucket.controller.ts
htuna07: Sure.

stacks/api/bucket/bucket.controller.ts
thesayyn: I think that checking for the `schema.type` is enough for this logic.

stacks/api/bucket/bucket.controller.ts
htuna07: Of course, they should be like you said. But it's a recursive method and I didn't want to take the risk of interrupting that logic with this kind of errors. Our client-side can't send this kind of mistaken schema if I'm not wrong?

thesayyn: Yes, it would send a schema whose properties property has '{}' as a value when there is no property. 

Also, note that ignoring a possible error may lead to unpredictable and unforeseen bugs.

htuna07: Okay, it looks better I remove this then, thanks for the suggestion.

stacks/api/bucket/utilities.ts
```diff
@@ -0,0 +1,68 @@
...
+  path = `${path}.$[]`;
+  if (isArray(previousSchema) && isArray(currentSchema)) {
+    addArrayPattern(previousSchema.items, currentSchema.items, removedKeys, path);
+  } else if (isObject(previousSchema) && isObject(currentSchema)) {
+    findRemovedKeys(previousSchema.properties, currentSchema.properties, removedKeys, path);
+  }
+}
+
+export function isObject(schema: any) {
+  return schema.type == "object" && Object.keys(schema.properties).length > 0 ? true : false;
```
thesayyn, line 59:
```suggestion
  return schema.type == "object" ;
```

stacks/api/bucket/utilities.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export function isRelation(schema: any, bucketId: string) {
+  return schema.type == "relation" && schema.bucketId == bucketId ? true : false;
+}
+
+export function isArray(schema: any) {
+  return schema.type == "array" ? true : false;
```
thesayyn, line 67:
```suggestion
  return schema.type == "array";
```

stacks/api/bucket/utilities.ts
htuna07: When the schema like `{ type:"object",properties:{} }`, it will be failed. I think this schema is valid but doesn't need to be checked for child keys.

thesayyn: Why would it fail if there are no properties?  Also, that is can't be a case since the user has to have at least one property. so it should be safe to remove that condition.

htuna07: Because if it's an object, it will pass properties of object to this recursive method and it will be failed at `for(const key of Object.keys(schema))`. Also, the user can send this kind of schema inside of object field. For example;
1- Create a new bucket and add an object field that doesn't have any property.

PR #294, opened 2020-05-06
style: design inconsistencies

htuna07:


stacks/spica/src/passport/pages/identity-add/identity-add.component.html
```diff
@@ -21,6 +21,7 @@ <h4>
         <mat-error *ngIf="identifierModel?.errors?.required">This field is required.</mat-error>
       </mat-form-field>
       <button
+        class="change-password"
```
thesayyn, line 24: In many cases, you don't need a class selector at all. Can accomplish the thing with a more strict CSS selector.

For instance:   mat-card-actions button:last:-of-type  instead of button.change-password

stacks/spica/src/passport/pages/strategies/strategies.component.scss
```diff
@@ -5,3 +5,10 @@
     width: 150px;
   }
 }
+
+::ng-deep {
```
thesayyn, line 9: This should be working without ::ng-deep  IMHO.

stacks/spica/packages/core/layout/home/home.layout.ts
```diff
@@ -83,6 +82,5 @@ export class HomeLayoutComponent implements OnInit {
 
   toggle(): void {
     this.expanded = !this.expanded;
-    setTimeout(() => this.sidenav.updateContentMargins(), 400);
```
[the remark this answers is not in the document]
htuna07: We handle all operations about the transition with pure css codes. So we don't need this line anymore.

PR #302, opened 2020-05-08
refactor: take options through arguments

thesayyn:
Please carefully review this one. I might have missed something.


stacks/api/main.ts
```diff
@@ -1,23 +1,184 @@
...
+    },
+    "database-replica-set": {
+      string: true,
+      alias: ["replica-set"],
+      description: "Name of the replica set."
+    },
+    "database-pool-size": {
+      number: true,
+      description: "Amount of connection that will be opened against database.",
+      default: 100
```
htuna07, line 48: It was supposed to be 10, right?

stacks/api/function/horizon/src/enqueuer.ts
```diff
@@ -1,7 +1,7 @@
 import {Enqueuer} from "@spica-server/function/enqueuer";
 import {EventQueue, Queue} from "@spica-server/function/queue";
 
-export const ENQUEUER = "HORIZON_SCHEDULER";
+export const ENQUEUER = Symbol.for("HORIZON_SCHEDULER");
```
htuna07, line 4: Why did we decide to use 'Symbol'?

stacks/api/main.ts
```diff
@@ -1,23 +1,184 @@
...
+    normalize: true,
+    description: "A path that will be used by API to store persistent data."
+  })
+  .option("public-url", {
+    string: true,
+    description: "Publicly accessible url of the APIs.",
+    alias: ["public-host"],
+    describe: `This option will be used by various modules. 
+For instance the 'storage' module will use this url to generate object links that are accessible from egress networks
+While the 'passport' module will use this url to re-route the user to the Passport APIs.
```
htuna07, line 117: Don't we need to put a default value for this public-url like 'http://localhost:4300'?

thesayyn: Our development defaults come from .bazelrc through the environment variables. So we don't need to put a default. instead, we demand this argument to be filled by the invoker.

stacks/api/function/horizon/src/enqueuer.ts
thesayyn: I wanted to prevent the misuse of injection tokens so I decided to use something like this.

thesayyn: In the previous code, you could use both `@Inject(HORIZON)` and `@Inject("HORIZON_SCHEDULER")` since the string type is a value type so both of them would work if I left the code as it was.

stacks/api/main.ts
thesayyn: I didn't notice that 👍 Let me change that

PR #303, opened 2020-05-08
feat(function): ability to filter channels

thesayyn:


stacks/api/function/horizon/src/horizon.ts
```diff
@@ -99,13 +99,13 @@ export class Horizon implements OnModuleInit {
     const path = event.target.cwd.split("/");
     const functionId = path[path.length - 1];
 
-    const stdOut = this.output.create({
+    const [stdout, stderr] = this.output.create({
       eventId: event.id,
       functionId
     });
 
     // TODO: forward stderr to another channel
```
htuna07, line 107: We can remove this comment line.

stacks/api/function/runtime/test/io/database.spec.ts
```diff
@@ -20,13 +20,29 @@ describe("IO Database", () => {
...
         content: "this is my message",
+        channel: "stdout",
+        event_id: "event",
+        function: "1"
+      });
+      done();
+    });
+  });
+
+  it("should write std to collection", done => {
```
htuna07, line 38: This text should be `should write stderr to collection`, isn't it?

thesayyn: Yes, it should be like that. Thanks

stacks/spica/src/function/pages/log-view/log-view.component.html
```diff
@@ -43,6 +43,9 @@ <h4 class="mat-h4">
       </mat-option>
     </mat-select>
   </mat-form-field>
+  <mat-slide-toggle [(ngModel)]="showErrors" (ngModelChange)="filter$.next(filter$.value)">
+    Hide Errors
```
htuna07, line 47: When this toggle switched true, it will display the errors but this text still stays as **Hide Error** which means the opposite of current behaviour. It should be **Show Errors** instead of **Hide Errors** or it should change when toggle switched.

PR #308, opened 2020-05-11
fix(bucket): store relations as string in database

htuna07:


packages/core/schema/defaults.ts
```diff
@@ -28,6 +28,17 @@ export const OBJECT_ID: Format = {
   }
 };
 
+export const OBJECTID_STRING: Format = {
+  name: "objectid-string",
+  type: "string",
+  coerce: bucketId => {
+    return bucketId;
+  },
```
thesayyn, line 36: You can remove the coerce function since it does nothing.

PR #310, opened 2020-05-11
ci: add public url for deployment 

htuna07:


thesayyn:
We should merge this one before others so @TolgaYigit can test it continuously. @htuna07

PR #315, opened 2020-05-15
feat(storage): storage upload response should include accessible url

htuna07:


stacks/api/storage/storage.service.ts
```diff
@@ -80,13 +80,11 @@ export class Storage {
...
-    Promise.all(
-      data.map(d => {
-        d._id = new ObjectId(d._id);
-        fs.writeFileSync(this.buildPath(d), d.content.data);
-        delete d.content.data;
-      })
-    );
+    data.forEach(d => {
+      d._id = new ObjectId(d._id);
+      fs.writeFileSync(this.buildPath(d), d.content.data);
```
thesayyn, line 85: For performance reasons can we replace this with fs.promises.writeFile instead? 
Cause we don't want to stop the event queue during IO operations.

PR #321, opened 2020-05-18
fix(bucket): column width handler affects the other columns

htuna07:


stacks/spica/packages/material/resize/resize.directive.ts
```diff
@@ -50,6 +51,13 @@ export class MatResizeHeader {
     @Optional() @Host() private matSort: MatSortHeader
   ) {}
 
+  ngAfterViewInit() {
+    setTimeout(() => {
+      this._width = parseInt(window.getComputedStyle(this.elem.nativeElement).width);
+      this.resizeend.emit(this._width);
```
thesayyn, line 57: It ought not to emit the "resize end" event since this is not a user-triggered resize.

PR #326, opened 2020-05-22
feat(passport): allow user to perform auto-login with JWT token

htuna07:


stacks/spica/src/passport/pages/identify/identify.component.ts
```diff
@@ -15,13 +16,26 @@ export class IdentifyComponent implements OnInit {
...
+        map(params => params.token),
+        tap(token => {
+          if (token) this.passport.token = token;
+          if (this.passport.identified) {
+            this.router.navigate([""]);
+          }
+        }),
+        take(1)
+      )
+      .subscribe();
```
thesayyn, line 38:
```suggestion
    this.activatedRoute.queryParams
      .pipe(
        filter(params => params.token)
        take(1)
      )
      .subscribe(params => {
           if (params.token) {
              this.passport.token = token;
           }
          if (this.passport.identified) {
            this.router.navigate([""]);
          }
      });
```
I think this one more appropriate for this logic.

PR #328, opened 2020-05-28
feat(function): kill functions when time out

htuna07:


stacks/api/function/horizon/test/horizon.spec.ts
```diff
@@ -66,19 +67,20 @@ describe("horizon", () => {
...
   let app: INestApplication;
   let spawnSpy: jasmine.Spy;
+  let horizonOptions = {
+    databaseUri: undefined,
+    databaseName: undefined,
+    databaseReplicaSet: undefined,
+    poolSize: 10,
+    publicUrl: undefined,
+    timeout: 60000
+  };
```
thesayyn, line 77: I couldn't get why you needed such a change.

stacks/api/main.ts
```diff
@@ -102,6 +102,13 @@ const args = yargs
       default: 10
     }
   })
+  .options({
+    "function-timeout": {
+      number: true,
+      description: "The timeout value to be reached before kill the process",
+      default: 60000
+    }
```
thesayyn, line 110: There a few things to correct here.

1- Can we just put this option into the same group with function-pool-size cause these options belong to function module.
2- Change description to: `Amount of time in milliseconds that has to elapse before aborting a function.`

stacks/api/function/horizon/test/horizon.spec.ts
htuna07: I need to pass the timeout value to this function `clock.tick();`

htuna07: https://github.com/spica-engine/spica/blob/9e01d3c3e20c52d3ef1619465e5aaf2b99b4a433/stacks/api/function/horizon/test/horizon.spec.ts#L155

PR #340, opened 2020-06-02
refactor(storage): remove close icon from storage object display

htuna07:


stacks/spica/src/storage/components/storage-dialog-overview/storage-dialog-overview.ts
```diff
@@ -12,8 +12,4 @@ export class StorageDialogOverviewDialog {
     public dialogRef: MatDialogRef<StorageDialogOverviewDialog>,
     @Inject(MAT_DIALOG_DATA) public data: any
```
thesayyn, line 13:
```suggestion
    @Inject(MAT_DIALOG_DATA) public data: any
```

You can remove the unused dialogRef property.

stacks/spica/src/storage/components/storage-dialog-overview/storage-dialog-overwiev.spec.ts
```diff
@@ -1,41 +0,0 @@
-import {HttpClientTestingModule} from "@angular/common/http/testing";
```
thesayyn, line 1: I'd expect new tests to be added.  
Can we add at least an "it block" that checks whether the data property is passed to the storage-view component?

PR #347, opened 2020-06-04
fix(function): bucket triggered function does not save and error message stays on each function

htuna07:

### DIFF
--- a/stacks/api/function/src/function.controller.ts
+++ b/stacks/api/function/src/function.controller.ts
@@ -102,7 +102,7 @@ export class FunctionController {
 
     return triggers
       .filter(trigger => trigger.type == "bucket")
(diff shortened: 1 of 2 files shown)


stacks/api/function/src/function.controller.ts
```diff
@@ -102,7 +102,7 @@ export class FunctionController {
 
     return triggers
       .filter(trigger => trigger.type == "bucket")
-      .some((trigger, index) => {
+      .some((trigger, index, triggers) => {
         const foundIndex = triggers.findIndex(
```
htuna07, line 106: `triggers.findIndex` checks each trigger of triggers array which is not filtered for only bucket triggers and it causes to find the wrong index. It must check only filtered functions to find the correct indexes.

PR #354, opened 2020-06-10
fix: various fixes for function and passport module

thesayyn:
This also fixes SP-348.


stacks/api/function/queue/node/http.ts
```diff
@@ -87,6 +87,10 @@ export class Request {
       this.params = new Map(req.params.map(h => [h.key, h.value]));
     }
 
+    if (req.query) {
+      this.query = JSON.parse(req.query);
+    }
```
htuna07, line 92: I think we should cover this `JSON.parse()` method with try-catch block. Otherwise, the backend would down when parse progress failed just like other JSON.parse lines caused.

stacks/api/function/src/function.controller.ts
```diff
@@ -228,7 +228,9 @@ export class FunctionController {
     }
     operators.push(
       last(),
-      switchMap(() => this.engine.compile(fn).catch(() => {})),
+      tap(() => {
+        this.engine.compile(fn).catch(() => {});
+      }),
```
htuna07, line 233: If we will make this change, the dependency install bug named SP-348 will be fixed, isn't it?

thesayyn: Yes.

stacks/api/function/queue/node/http.ts
thesayyn: Unlike other errors, this one would appear in the logs. If we catch it with a try, how would we handle it? I assume we would re-throw this error as is.

PR #357, opened 2020-06-12
feat(function): introducing the bucket before delete trigger

htuna07:


stacks/api/bucket/bucket-data.controller.ts
```diff
@@ -28,6 +29,7 @@ import * as locale from "locale";
 import {createBucketDataActivity} from "./activity.resource";
 import {BucketDataService, getBucketDataCollection} from "./bucket-data.service";
 import {findRelations} from "./utilities";
+import {createAuthGuard} from "@spica-server/passport/auth.guard";
```
thesayyn, line 32: Unused

PR #362, opened 2020-06-18
feat(passport): add support for strategy query param to initiate authentication

thesayyn:


charts/spica/templates/database.yaml
```diff
@@ -13,10 +20,16 @@ spec:
...
+        app.kubernetes.io/component: database
+        app.kubernetes.io/instance: {{ .Release.Name }}
+        app.kubernetes.io/version: {{ .Chart.AppVersion }}
+        app.kubernetes.io/part-of: {{ .Chart.Name }}
+        app.kubernetes.io/managed-by: helm
     spec:
       initContainers:
         - name: replication-controller
-          image: spicaengine/mongoreplicationcontroller:{{ $.Chart.AppVersion }}
+          image: spicaengine/mongoreplicationcontroller:0.3.1
```
htuna07, line 32: Why did we decide to make this change?

thesayyn: It is mistaken. Good catch

PR #363, opened 2020-06-18
fix(passport): do not allow to user to login with empty password

htuna07:


stacks/api/passport/passport.service.ts
```diff
@@ -47,6 +47,9 @@ export class PassportService {
   }
 
   identify(identifier: string, password: string): Promise<Identity | null> {
+    if (password == undefined || password == null || password == "") {
+      return new Promise(resolve => resolve(null));
+    }
```
thesayyn, line 52:
```suggestion
    if (password == undefined || password == null || password == "") {
      return Promise.resolve(null);
    }
```

thesayyn: Could be better.

PR #367, opened 2020-06-22
fix(bucket): scheduled entries are not working

htuna07:


thesayyn:
Can you at least provide some test cases so we can catch this kind of issue in the future?

PR #382, opened 2020-07-03
feat(passport): support for specific resourced policy statements

htuna07:


stacks/api/passport/guard/test/action.guard.spec.ts
```diff
@@ -0,0 +1,304 @@
...
+      params: {
+        id: "test1",
+        row: "test2"
+      },
+      statements: [
+        {
+          action: "bucket:data:show",
+          resource: {
+            include: "test/test1",
+            exclude: ["test2"]
```
htuna07, line 276: We should send excluded parameters like 'test/test2'

stacks/api/passport/guard/test/action.guard.spec.ts
```diff
@@ -0,0 +1,304 @@
...
+            include: "test/test3"
+          },
+          module: "bucket:data"
+        }
+      ],
+      resourceFilter: true
+    });
+    await guard();
+
+    // We expect the wildcard to not appear in the include array
```
htuna07, line 161: This expectation should be include wildcard i think.

thesayyn (approved):

PR #384, opened 2020-07-09
feat(function): support for function live logs

htuna07:


stacks/spica/src/function/function.service.ts
```diff
@@ -34,19 +41,35 @@ export class FunctionService {
...
     }
-    return this.http.get<any[]>(`api:/function/logs`, {params: serializedFilter});
+
+    queryParams.push(`limit=${limit}`);
+    queryParams.push(`skip=${skip}`);
+    queryParams.push(`sort=${JSON.stringify(sort)}`);
+
+    queryParams.push(`Authorization=${this.passport.token}`);
+
+    let mergedParams = "?" + queryParams.join("&");
```
thesayyn, line 70: You could use the URL class to do this which is more appropriate to build up links like this.

PR #396, opened 2020-07-17
docs: update dashboard content

TolgaYigit:


docs/content/guide/dashboard.md
```diff
@@ -144,41 +144,86 @@ export default function(req, res) {
...
-```typescript
+```javascript
 // Add `@spica-devkit/dashboard` to your dependencies and import it
 import * as Dashboard from "@spica-devkit/dashboard";
-export default async function () {
+export default async function (req,res) {
   // Create an API key and initialize dashboards with it
-  Dashboard.initialize(<APIKEY>);
+
+  Dashboard.initialize("1i7z1kcoxehci");
```
thesayyn, line 169: I am not sure if this is the current way of initializing the dashboard package.
@htuna07 knows better about this. Do you think that this code is still valid?

Also, can we replace the API key with a placeholder one? like so
```suggestion
  Dashboard.initialize("<YOUR_API_KEY_GOES_HERE>");
```
so that the user doesn't assume that this code works out of the box.

docs/content/guide/dashboard.md
```diff
@@ -144,41 +144,86 @@ export default function(req, res) {
...
     components: [
+        {
+        type: "bar",
+        // Url of your dashboard function
+        url: "https://master.spicaengine.com/api/fn-execute/line",
+        key: "bardata",
+      }, 
+      {
+        type: "table",
+        // Url of your dashobard function
```
thesayyn, line 184: Maybe we should not explicitly use the functions as the data source so that the user doesn't assume that data source URLs have to be a function with HTTP trigger. 

For instance, we could fetch data from a weather forecast API?

docs/content/guide/dashboard.md
htuna07: We updated this initialize method like `Dashboard.initialize({apikey:"1i7z1kcoxehci"})` from version 0.5.0(latest) of @spica-devkit/dashboard package.

thesayyn: @TolgaYigit then we have to update the documentation as well.

thesayyn:
Hello @TolgaYigit. ☺️☺️☺️☺️☺️  

This PR is going to be an historical artifact if we wait some more.

thesayyn (approved):

PR #397, opened 2020-07-17
fix(common): make array input more robust

htuna07:
This pr includes SP-407 and 408. Also there are some changes on returned error message from schema validation.


thesayyn:
Fixes SP-408 and SP-407

PR #405, opened 2020-07-24
feat(core): allow to start api with cors parameters

htuna07:


packages/core/middlewares.ts
```diff
@@ -19,26 +21,53 @@ export namespace Middlewares {
...
-    if (req.header("access-control-request-method")) {
-      res.header("Access-Control-Allow-Methods", req.header("access-control-request-method"));
-    }
+      res.header("Access-Control-Allow-Origin", allowedOrigin);
 
-    let allowedHeaders = "Authorization, Content-Type, Accept-Language";
-    if (req.header("access-control-request-headers")) {
-      allowedHeaders = req.header("access-control-request-headers");
-    }
+      //console.log(res.header("Access-Control-Allow-Origin"));
```
thesayyn, line 30: You can remove this line.

packages/core/middlewares.ts
```diff
@@ -19,26 +21,53 @@ export namespace Middlewares {
...
-    } else {
-      next();
-    }
+      let allowedHeaders = options.allowedHeaders.join(",");
+
+      if (req.header("access-control-request-headers")) {
+        allowedHeaders = getMatchedValue(
+          req.header("access-control-request-headers").indexOf(",") != -1
+            ? req.header("access-control-request-headers").split(",")
+            : req.header("access-control-request-headers"),
```
thesayyn, line 47:
```suggestion
          req.header("access-control-request-headers").split(",")
```
Could we just use the array all the time? I mean if there is no comma char in the string this logic will return an array whose first item's content is the content of the header. So I think may not a logic like that

PR #409, opened 2020-07-28
fix(storage): show upload limit on error message

htuna07:
Somehow error.message is undefined on master.spicaengine


stacks/api/storage/src/body.ts
```diff
@@ -27,7 +27,12 @@ abstract class __BsonBody {
       parser(req, res, error => {
         if (error) {
           if (error.type == "entity.too.large") {
-            return observer.error(new HttpException(error.message, HttpStatus.PAYLOAD_TOO_LARGE));
+            return observer.error(
+              new HttpException(
+                error.message + ` maximum file size is ${this.options.objectSizeLimit} MB `,
```
thesayyn, line 32: If error.message is undefined on prod then we should not depend on it at all since it could make confusing messages if it ever works again.

```suggestion
           `maximum object size is ${this.options.objectSizeLimit}Mi`,
```

PR #434, opened 2020-08-21
docs(api): improve api docs

thesayyn:


stacks/api/bucket/src/bucket-data.controller.ts
```diff
@@ -237,16 +269,41 @@ export class BucketDataController {
...
+   *    "age": 35,
+   *    "famous": true,
+   *    "role": "Actor"
+   * }
+   * ```
+   * ##### Otherwise, when there is a translated property, for instance the `name` property has to be like below
+   * ```json
+   * {
+   *    "name": {
+   *      "tr_TR": "Lusifır",
```
TolgaYigit, line 289: Probably we can find a better and simple translation example. Maybe a book entry?

TolgaYigit (changes requested):

stacks/api/bucket/src/bucket-data.controller.ts
thesayyn: Yes, that makes sense. Do you have anything in mind?  
It would be helpful if you can suggest a change

thesayyn:
@TolgaYigit @htuna07 ping

PR #435, opened 2020-08-25
refactor: change left-hand menu icons and move settings page into identity

thesayyn:


TolgaYigit (approved):
Quality Control Approved 👍🥇

stacks/spica/src/styles.scss
```diff
@@ -58,14 +58,25 @@ body {
...
+    justify-content: flex-end;
+    .save-button {
+      mat-icon {
+        margin-right: 5px;
+      }
+      mat-progress-spinner {
+        display: inline-block;
+        margin-right: 5px;
+      }
+    }
```
htuna07, line 78: After these changes, save buttons which is not inside of mat-card-actions still stays on the left of content. You can check some passport pages. We can cover save buttons with mat-card-actions.

htuna07 (approved):

PR #449, opened 2020-09-09
feat(bucket): update bucket entry when bucket schema field type updated

htuna07:


thesayyn:
This logic gets so confusing and unmaintainable. I guess it is time to use the differ logic that we have used in bucket/history which perfectly works for history. I guess it'd work perfectly for this job too.

I'll lift it up after the stack implementation which will require us to do some refactoring in various places and this is one them.

thesayyn (approved):

PR #453, opened 2020-09-18
feat(bucket): update bucket entry when bucket language removed

htuna07:


stacks/api/bucket/src/bucket.module.ts
```diff
@@ -75,6 +78,21 @@ export class BucketModule {
...
+    BucketDataService,
+    {
+      provide: LANGUAGE_CHANGE_UPDATER,
+      useFactory: provideLanguageChangeUpdater,
+      inject: [BucketService, BucketDataService]
+    }
+  ],
+  exports: [LANGUAGE_CHANGE_UPDATER]
+})
+export class PreferencesListenerModule {}
```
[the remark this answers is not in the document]
thesayyn: to follow up the convention.

stacks/api/bucket/src/bucket.schema.resolver.ts
```diff
@@ -1,30 +1,35 @@
...
-            format: "date-time"
-          };
-        }
-        return schema || {type: true};
-      });
-    }
+  resolve(uri: string): Observable<object> {
+    return combineLatest(
+      this.preferenceWatcher,
+      from(this.bucketService.findOne({_id: new ObjectId(uri)}))
```
thesayyn, line 19: How about we make a method in the bucket service that returns an observable which will emit the latest version of an arbitrary bucket whenever it changes? 

Therefore we could get rid of BucketCache once for all?

thesayyn: If we make these suggested changes, we wouldn't need the BucketCache cause this class would both invalidate either of these (preference or bucket itself) changes.

stacks/api/bucket/src/utility.ts
```diff
@@ -170,3 +172,72 @@ export function buildRelationAggregation(
     ];
   }
 }
+
+export function provideLanguageChangeUpdater(
+  bucketService: BucketService,
+  bucketDataService: BucketDataService
+) {
+  return (previousSchema: object, currentSchema: object) => {
```
thesayyn, line 180: How about we make the returned function an async function? I believe if we get rid of .then chaining, it'd be more readable.

This is only a suggestion. You might like it this way. NP

thesayyn (changes requested):

thesayyn (commented):
Looks better. Good work! 🥳

thesayyn (approved):

PR #457, opened 2020-09-23
feat(bucket): read-only option for bucket fields

htuna07:


stacks/api/bucket/src/bucket.schema.resolver.ts
```diff
@@ -38,8 +38,43 @@ export class BucketSchemaResolver {
...
+  let keywords = [
+    {
+      name: "default",
+      def: {
+        modifying: true,
+        compile: (schema, parentSchema, it) => {
+          return (data, dataPath, parentData) => {
+            console.log(schema, parentSchema);
+
+            console.log(data, dataPath, parentData);
```
thesayyn, line 59: You have forgotten to remove these lines.

thesayyn (changes requested):

thesayyn (approved):

thesayyn:
I want to ask something however. Can I use readonly without default value?

htuna07:
> I want to ask something however. Can I use readonly without default value?

Yes, you can. But the default value will be `undefined` in this case.

htuna07:
> > I want to ask something however. Can I use readonly without default value?
> 
> Yes, you can. But the default value will be `undefined` in this case.

Actually, it won't work without default keyword, sorry.

PR #458, opened 2020-09-23
fix(bucket): generate bucket guide requests safely

htuna07:


stacks/spica/src/bucket/pages/index/index.component.html
```diff
@@ -174,7 +174,7 @@ <h5>
             mat-resize-header
             [persist-header-width]="schema._id"
           >
-            id
+            _id
```
thesayyn, line 177:
```suggestion
            id
```

The end-user does not need to know that this field has leading underscore. Can we revert this change?

stacks/spica/src/bucket/pages/index/index.component.ts
```diff
@@ -83,7 +83,7 @@ export class IndexComponent implements OnInit {
         this.guideResponse = {};
         this.readOnly = schema.readOnly;
         this.properties = [
-          {name: "$$spicainternal_id", title: "id"},
+          {name: "$$spicainternal_id", title: "_id"},
```
thesayyn, line 86: Same as above

thesayyn (changes requested):

stacks/spica/src/bucket/pages/index/index.component.html
htuna07: User can create a field  named as **id** but not **_id**. If user creates **id** field there will be 2 id field and it will be more confusing. Also guide responses include **_id** instead of **id**. I think these changes will make things more understandable for users.

thesayyn: Yes that makes sense

thesayyn (approved):

PR #462, opened 2020-10-02
fix(passport): handle x-www-form-urlencoded payloads in assertion end…

thesayyn:
…point


stacks/spica/src/function/function.initializer.ts
```diff
@@ -36,9 +36,11 @@ export class FunctionInitializer {
 
   async appInitializer() {
     const allowed = await this.passport.checkAllowed("function:index", "*").toPromise();
+    console.log(allowed);
     if (this.passport.identified && allowed) {
       this.functionService.loadFunctions().toPromise();
     } else {
+      console.log(allowed);
```
htuna07, line 43: We should remove these console.log lines

htuna07 (approved):

htuna07 (approved):

PR #471, opened 2020-10-19
feat(function): added sidecar log view

akhnos:


stacks/spica/src/function/pages/add/add.component.scss
```diff
@@ -1,14 +1,33 @@
...
+      margin-bottom: 60px;
+    }
+
+    div.actions {
+      max-width: 100%;
+      align-self: flex-end;
+      overflow-x: scroll;
+
+      -ms-overflow-style: none; /* Internet Explorer 10+ */
+      scrollbar-width: none; /* Firefox */
```
thesayyn, line 27: These are redundant since we have autoprefixer built into Angular CLI. therefore these lines should be removed or replaced with a standard CSS rule.  

See https://github.com/spica-engine/spica/blob/master/stacks/spica/browserslist

stacks/spica/src/function/pages/add/add.component.scss
```diff
@@ -1,14 +1,33 @@
...
+      &::-webkit-scrollbar {
+        display: none; /* Safari and Chrome */
```
thesayyn, line 29:
```suggestion
        display: none;
```

stacks/spica/src/function/pages/log-view/log-view.component.html
```diff
@@ -3,10 +3,17 @@ <h4 class="mat-h4">
...
+    <button
+      mat-button
+      canInteract="function:logs:delete"
+      (click)="clearLogs()"
+      [disabled]="!(queryParams | async)?.function.length"
+    >
+      <mat-icon>clear_all</mat-icon>
+      Clear
+    </button>
+  </div>
```
thesayyn, line 16: Why we had to wrap this button with a div? If this had been done to use a tooltip, you can use it directly on the button.

akhnos: Because tooltip is disabled when button is disabled. Tooltip should work even if the button is disabled.

htuna07 (dismissed):

htuna07 (approved):

PR #472, opened 2020-10-19
feat(bucket): graphql support for bucket data

htuna07:


stacks/api/bucket/src/graphql/graphql.ts
```diff
@@ -0,0 +1,609 @@
...
+      }
+
+      let aggregation = [];
+
+      let requestedFields = requestedFieldsFromInfo(info);
+      let relationAndLocalization = await aggregationsFromRequestedFields(
+        JSON.parse(JSON.stringify(bucket)),
+        requestedFields,
+        this.getLocale,
+        JSON.parse(JSON.stringify(this.buckets))
```
thesayyn, line 471: Fetching all these buckets upfront and (un)serializing them is intense work.  
We should fetch buckets on an as-needed basis to lift off the pressure from the database and memory.

thesayyn: This has been done in several places. We have to fix them as well.

stacks/api/bucket/src/graphql/schema.ts
```diff
@@ -0,0 +1,559 @@
...
+    }
+
+    expressions[key] = value;
+  });
+
+  return expressions;
+}
+
+function findValueOfPath(path: (string | number)[], document: BucketDocument) {
+  return path.reduce((acc, curr) => acc[curr], document);
```
thesayyn, line 524:
```suggestion
  return path.reduce((document, name) => document[name], document);
```

stacks/api/bucket/src/graphql/schema.ts
```diff
@@ -0,0 +1,559 @@
...
+  return requestedFields.some(fields => relatedFields.includes(fields[0]));
+}
+
+function translatableFieldRequested(properties: object, requestedFields: string[][]): boolean {
+  let translatableFields = [];
+  Object.keys(properties).forEach(key => {
+    if (properties[key].options && properties[key].options.translate) {
+      translatableFields.push(key);
+    }
+  });
```
thesayyn, line 544: Use for in instead.

stacks/api/bucket/src/graphql/schema.ts
```diff
@@ -0,0 +1,559 @@
...
+
+  return requestedFields.some(fields => translatableFields.includes(fields[0]));
+}
+
+export function requestedFieldsFromExpression(expression: object, requestedFields: string[][]) {
+  for (const [key, value] of Object.entries(expression)) {
+    if (key == "$or" || key == "$and") {
+      value.forEach(condition => requestedFieldsFromExpression(condition, requestedFields));
```
thesayyn, line 552: Use for of instead.

thesayyn (changes requested):

thesayyn (approved):

PR #477, opened 2020-10-28
feat(function): show example code for trigger types

htuna07:


stacks/spica/src/function/examples/examples.json
```diff
@@ -0,0 +1,28 @@
+{
```
thesayyn, line 1: Can we use a typescript file instead of json and format these examples beforehand?

stacks/spica/src/function/pages/add/add.component.ts
```diff
@@ -24,8 +24,13 @@ import {
...
   Information,
   NormalizedFunction,
-  normalizeFunction
+  normalizeFunction,
+  Trigger
 } from "../../interface";
+import {MatDialog} from "@angular/material/dialog";
+import {CodeComponent} from "./code/code.component";
+import * as ts from "prettier/parser-typescript";
+import {format} from "prettier/standalone";
```
thesayyn, line 33: Importing these modules within the main bundle causes a significant bundle size growth. So we can not do something like this. 

I think the way to go is by doing something we'd done in https://github.com/spica-engine/spica/blob/master/stacks/spica/src/function/components/editor/snippets.ts.

thesayyn (changes requested):

thesayyn (approved):

PR #483, opened 2020-11-03
fix(bucket): do not add duplicated enum values to graphql schema

htuna07:


stacks/api/bucket/src/graphql/schema.ts
```diff
@@ -88,11 +88,15 @@ function createInterface(
 function createEnum(name: string, values: string[]) {
   return `
       enum ${name}{
-        ${values.join("\n")}
+        ${removeDuplicates(values).join("\n")}
```
thesayyn, line 91:
```suggestion
        ${new Set(values).join("\n")}
```

thesayyn (changes requested):

stacks/api/bucket/src/graphql/schema.ts
htuna07: It works with `Array.from(new Set(values).values()).join("\n")`

thesayyn (approved):

PR #485, opened 2020-11-04
refactor: font-size changed, dark mode changed

akhnos:


stacks/spica/src/styles.scss
```diff
@@ -103,3 +106,7 @@ mat-form-field {
     background-color: #fafafa !important;
   }
 }
+
+.mat-list-base .mat-list-item {
+  font-size: 1rem;
+}
```
thesayyn, line 112: We can not put this global style here. It should be in the component scope. If we leave this here, it will affect places that it not supposed to.

thesayyn (changes requested):

stacks/spica/src/styles.scss
akhnos: It is an intended change. We actually changed the theme settings and override font-size:16px.

thesayyn: Then could you do it by using the material's typography mixins?

https://material.angular.io/guide/typography

akhnos: Change requests accepted

thesayyn (approved):

PR #486, opened 2020-11-05
fix(storage): add timestamp to storage url after content updated

htuna07:


stacks/spica/src/storage/pages/index/index.component.ts
```diff
@@ -52,12 +52,29 @@ export class IndexComponent implements OnInit {
...
+              if (!updatedId) {
+                return storages;
+              }
+
+              storages.data = storages.data.map(s => {
+                if (s._id == updatedId) {
+                  s.url = s.url + "?timestamp=" + new Date().getTime();
+                }
+                return s;
+              });
```
thesayyn, line 73: Can you implement this logic via findIndex ?

thesayyn (changes requested):

thesayyn (approved):

PR #492, opened 2020-11-09
fix(bucket): make bucket id and relation type required for relation property

htuna07:


stacks/spica/src/bucket/components/relation-schema/relation-schema.component.ts
```diff
@@ -1,19 +1,36 @@
...
+      if (!this.schema.relationType) {
+        this.schema.relationType = RelationType.OneToOne;
+      }
+      this.buckets = this.bucketService.getBuckets().pipe(
+        tap(buckets => {
+          if (!this.schema.bucketId) {
+            this.schema.bucketId = buckets[0]._id;
+          }
+        })
+      );
```
thesayyn, line 33: You can put these into constructor or ngOnInit. then you wouldn't need to use setTimeout.

htuna07: **Expression has changed after it was checked** error shows up when I made these changes.

thesayyn: How about doing it in the constructor?

htuna07: Same error.

htuna07: It works in the constructor with setTimeout.

thesayyn (approved):

PR #494, opened 2020-11-10
fix(passport): update identities after identity settings updated

htuna07:


stacks/api/passport/identity/src/utility.ts
```diff
@@ -8,3 +10,83 @@ export function attachIdentityAccess(request: any) {
...
+      let fieldName = "attributes." + current;
+      acc = {...acc, [fieldName]: ""};
+      return acc;
+    }, {});
+
+    return identityService.updateMany({}, {$unset: unsetFields});
+  };
+}
+
+export function findUpdatedFields(
```
htuna07, line 35: @thesayyn I know these codes are copied from bucket utilities and it should not be duplicated. Where should we put these common codes which are common for only bucket and identity?

thesayyn: Can we hold this PR this we land bucket rules? only then I could think of something to about this.

thesayyn:
Now that we have the differ extracted to the common package, can we make use of it to get rid of custom differ logic?

thesayyn (approved):

thesayyn (approved):

PR #497, opened 2020-11-11
fix(bucket): prevent some names and enums

htuna07:


thesayyn:
I don't think that changing the APIs to match GraphQLs specification is a good idea.

Here's what we could do about those restrictions;

1. Ignore integer enums -> instead accept any number and validate them in the validation layer. The same thing applies to string enums as well.
2. Generate warnings about field names that do not match GraphQLs specification and ignore them gracefully.
3. Currently, the API strictly checks if the "relationType" and "bucketId" fields are present. All we have to do is check if the related bucket is missing and if it is then generate a warning about it.

thesayyn:
The idea behind these changes is;

It's GraphQL's limitation, not Spicas

PR #502, opened 2020-11-17
fix(function): function triggers should be unregistered after deletion

htuna07:


thesayyn:
LGTM

thesayyn (approved):

thesayyn (changes requested):
overall looks good to me but could we, use `const` where it is possible? @htuna07

thesayyn:
also when we land this one, we can land other PRs that want to make use of the differ.

thesayyn (approved):

PR #516, opened 2020-11-27
refactor(bucket): merge graphql and controller CRUD actions

htuna07:


stacks/api/bucket/src/crud.ts
```diff
@@ -0,0 +1,285 @@
...
+
+  const aggregations = [];
+
+  // resourcefilter
+  if (Object.keys(params.resourceFilter || {}).length) {
+    aggregations.push(params.resourceFilter);
+  }
+
+  // scheduled contents
+  aggregations.push({$match: {_schedule: {$exists: !!options.schedule}}});
```
thesayyn, line 47:
```suggestion
  if ( !options.schedule ) {
       aggregations.push({$match: {_schedule: {$exists: false}}});
  }
  
```

thesayyn: since we want to keep aggregations as small as possible, this would be more appropriate

stacks/api/bucket/src/bucket-data.controller.ts
```diff
@@ -325,40 +204,22 @@ export class BucketDataController {
...
+      schema,
+      rawDocument,
+      {req: req},
+      {
+        collection: bucketId => this.bds.children(bucketId),
+        schema: (bucketId: string) => this.bs.findOne({_id: new ObjectId(bucketId)})
+      }
+    ).catch(error => {
+      throw error;
     });
```
thesayyn, line 217:
```suggestion
    );
```

thesayyn: seems redundant

stacks/api/bucket/src/bucket-data.controller.ts
```diff
@@ -489,39 +333,24 @@ export class BucketDataController {
...
+      patch,
+      {req: req},
+      {
+        collection: bucketId => this.bds.children(bucketId),
+        schema: (bucketId: string) => this.bs.findOne({_id: new ObjectId(bucketId)})
+      },
+      {returnOriginal: false}
+    ).catch(error => {
+      throw error;
     });
```
thesayyn, line 348: Same as above. I am not sure if this was intentional but this is the same as letting the original promise reject.

thesayyn (changes requested):
Just some small improvement. Overall looks good. 👏 

It is good to see all of these redundant codes gone

stacks/api/bucket/src/bucket-data.controller.ts
htuna07: removed

stacks/api/bucket/src/bucket-data.controller.ts
htuna07: removed

thesayyn (approved):

PR #523, opened 2020-12-04
refactor(bucket): optimization for ordering relations

htuna07:


stacks/api/bucket/src/crud.ts
```diff
@@ -126,6 +155,16 @@ export async function findDocuments(
   return collection.aggregate([...aggregations, ...seekingPipeline]).toArray();
 }
 
+// it will work on only basic filters => {"user.name" : "John"}
+function extractFilterPropertyMap(filter: any) {
```
thesayyn, line 159: Keep in mind that this won't work for expressions too

thesayyn (approved):

thesayyn (approved):

PR #530, opened 2020-12-10
fix(bucket): update relation picker request

htuna07:


thesayyn (approved):
LGTM. 

keep in mind that you should change this we land ObjectId macro to the expression system.

PR #534, opened 2020-12-12
fix(function): update realtime logs due to filter expression changes

htuna07:
Additional changes on @spica-devkit/bucket


thesayyn (approved):
this should have been done with the previous relevant change. i am sorry that i missed this at the time

TolgaYigit (approved):

PR #538, opened 2020-12-15
fix(bucket): update aggregation stages to work with _id filter

htuna07:


stacks/api/bucket/src/crud.ts
```diff
@@ -87,6 +87,15 @@ export async function findDocuments<T>(
...
 
+  // matching for rules and filters
+  pipeline.push({
+    $set: {
+      _id: {
+        $toString: "$_id"
+      }
+    }
+  });
+
```
thesayyn, line 98: Keep in mind that this might cause `findOne` to do a `COLL_SCAN` because by doing this you are discarding the indexes on _id.

thesayyn (approved):
I am approving this by assuming that you know the consequences of _id coercion.

thesayyn (dismissed):

thesayyn (approved):

thesayyn (approved):

PR #542, opened 2020-12-17
refactor(function): decouple function runtime from api runtime

thesayyn:


thesayyn:
@htuna07 if you are willing to rebase this, i would like to see this landed as this is an important piece of the function module.

htuna07:
> @htuna07 if you are willing to rebase this, i would like to see this landed as this is an important piece of the function module.

I should make the failed CLI e2e tests pass before merging it.

PR #563, opened 2020-12-31
fix(function): async processes keep workers alive even it exceeds timeout

akhnos:


stacks/api/function/scheduler/src/scheduler.ts
```diff
@@ -246,7 +246,8 @@ export class Scheduler implements OnModuleInit, OnModuleDestroy {
       `an event has been completed ${id} with status ${succedded ? "success" : "fail"}`
     );
     this.processsingQueue.delete(id);
-    clearTimeout(this.timeouts.get(id));
+
```
htuna07, line 249:
```suggestion
// async processes keep workers alive even it exceeds timeout
```

htuna07 (approved):

PR #611, opened 2021-02-17
feat(storage): delete multiple storage objects on client

htuna07:


SpicaEngine (changes requested):
Instead of inline styles, we can use [class.active]="..."

akhnos (approved):

SpicaEngine (approved):

PR #803, opened 2022-01-11
refactor: migrate rules_nodejs to latest version

thesayyn:
PTAL.

This PR upgrades rules_nodejs to the latest version with a lot of changes. Ideally, there shouldn’t be any breaking changes since this PR changes the way the source is built. There are still a few failing tests due to glitches in how ts_library worked. Unfortunately, I don’t have the bandwidth to fix all failing tests so it would be great if @htuna07 you could fix the remaining failing tests.
 
- it upgrades nodejs version to 16.0.0 (this might break functions that target lower node runtime versions)
- rules_nodejs is upgraded to the latest version
- stacks/cli is separated from the rest of the stacks (i will send follow up PR to split up stacks/spica and stacks/api as well)
- usage of ts_library is completely dropped now ts_project is used for building

hit me up if you have a questions.


thesayyn:
PS; it should be really straightforward to fix the remaining tests since there are no architectural changes involved.

htuna07:
Thank you @thesayyn for these huge changes. I will handle the failing tests and other failed parts if any. By the way, could you explain the **separating cli from rest of the stacks**, please? It is supposed to be a separated part just like stacks/api and stacks/spica.

thesayyn:
With the new version of rules nodejs package hoisting is kinda broken. So I thought it would be the best to have separate package.jsons for each stack.

stacks/api/BUILD.bazel
```diff
@@ -1,11 +1,12 @@
+load("//tools/ts_project:index.bzl", "ts_project")
+
 package(default_visibility = ["//visibility:public"])
 
-load("@npm//@bazel/typescript:index.bzl", "ts_library")
 load("@build_bazel_rules_nodejs//:index.bzl", "nodejs_binary")
 load("@io_bazel_rules_docker//nodejs:image.bzl", "nodejs_image")
-load("@io_bazel_rules_docker//container:container.bzl", "container_image", "container_push")
+load("@io_bazel_rules_docker//container:container.bzl", "container_image")
```
htuna07, line 7: Was removing container_push an intended change?

thesayyn: Nope. It's supposed to be there. i guess i removed it when i ran buildozer

thesayyn:
Also, beware of changes in node runtime. I have removed `--experimental-loader` because it was no longer working.

thesayyn:
@htuna07  LMK if you need any help pushing this forward.

htuna07:
@thesayyn CLI tests are failing due to @bazel/jasmine version. Here is the output of the error; 

```
Test output for //stacks/cli/test/commands/bucket:schema_test:
  | [jasmine_runner.js] An error has been reported: TypeError: jrunner.onComplete is not a function
  | at main (node_modules/@bazel/jasmine/jasmine_runner.js:119:11)
  | at node_modules/@bazel/jasmine/jasmine_runner.js:182:32
  | at Object.<anonymous> (node_modules/@bazel/jasmine/jasmine_runner.js:187:5)
  | at Module._compile (node:internal/modules/cjs/loader:1108:14)
  | at Object.Module._extensions..js (node:internal/modules/cjs/loader:1137:10)
  | at Module.load (node:internal/modules/cjs/loader:988:32)
  | at Function.Module._load (node:internal/modules/cjs/loader:828:14)
  | at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:76:12)
  | at node:internal/main/run_main_module:17:47

```
Actually, it's supposed to use its own defined @bazel/jasmine version under the stacks/cli package.json but probably it doesn't. Have you ever faced this error?

thesayyn:
> Actually, it's supposed to use its own defined @bazel/jasmine version under the stacks/cli package.json but probably it doesn't. Have you ever faced this error?

ah yes. The jasmine version has to be downgraded. probably using the same version specified in the root package.json would help.

htuna07 (approved):

htuna07:
@thesayyn A little reminder :) I realized **nohoist** doesn't work with its current definition. It should be something like this `nohoist: ["**"]` if we want to disable hoisting.

thesayyn:
I didn’t even notice tbh. thanks for the info though

PR #846, opened 2022-04-21
fix(realtime): parse resource filters and apply

htuna07:


emre2345:
@htuna07 can you elaborate the pr please.

htuna07:
@emre2345 Somehow, our policy implementation is forgotten to add to the bucket-data realtime. 
The expectation after this PR is to be filtered returned bucket-data from bucket-data real-time connection if there are bucket-data stream policies.
1- Go to policies page and create a policy which filters bucket-data realtime documents. Use bucket-data stream actions for it. Attach that policy to the relevant apikey.
2- Connect bucket-data endpoint via websocket, use that apikey for authorization.
3- You are supposed to see the returned bucket documents from realtime connection are filtered depending on your policy specification.

emre2345 (approved):
LGTM

Tested on my local spica instance

PR #847, opened 2022-04-26
fix(passport): filter 2FA enabled identities with query

htuna07:


emre2345:
@htuna07 can you please elaborate the pr and test cases

htuna07:
@emre2345 We only changed the way how we filter identities that have enabled 2FA. Previously we filter them on the API, which causes API to down for the instances that have a large amount of identity, now we filter them on the MongoDB level with a query that is more optimized than the previous way.

htuna07:
@emre2345 What should be done to test:
1- Insert approximately 1,5M identities into the identities collection(I can share a script with you for this purpose)
2- Start the API and  you will see the memory usage of NodeJS will increase and an OOM error will be occurred at the end
3- This PR is supposed to prevent this OOM

emre2345:
@htuna07  I could not reproduce this on my end.

emre2345 (approved):
lgtm

PR #848, opened 2022-04-26
feat(versioncontrol): able to work with different function envs

htuna07:


emre2345 (commented):
Can you please elaborate the interfaces in this file:
stacks/api/versioncontrol/src/interface.ts

I could not understand what they're meant to do by looking into their names.

emre2345 (approved):
LGTM

PR #861, opened 2022-07-13
feat: schema grouping

Davuttrg:


stacks/spica/packages/core/layout/home/home.layout.ts
```diff
@@ -19,7 +19,7 @@ export class HomeLayoutComponent implements OnInit {
 
   expanded = true;
 
-  routes$: Observable<Route[]>;
+  routes$: Observable<[Route[]]>;
```
htuna07, line 22: Can we use a single object to keep routes with their categories instead of using an array of route array?

stacks/spica/src/bucket/pages/bucket-add/bucket-add.component.ts
```diff
@@ -70,10 +73,13 @@ export class BucketAddComponent implements OnInit, OnDestroy {
     private dialog: MatDialog
   ) {
     this.inputTypes = _inputResolver.entries();
-    this.bs
-      .getBuckets()
-      .pipe(take(1))
-      .subscribe(buckets => (this.buckets = buckets));
+    this.bsSubscription = this.bs.getBuckets().subscribe(buckets => {
```
htuna07, line 76: You can use filter operator with take(1) instead.

stacks/spica/packages/common/categorizing/categorizing.component.ts
```diff
@@ -0,0 +1,58 @@
+import {Component, OnInit, Input, Output, EventEmitter, SimpleChanges} from "@angular/core";
+import {MatDialog, MatDialogRef} from "@angular/material/dialog";
+
+@Component({
+  selector: "categorizing",
```
htuna07, line 5: Can we rename this component as category?

stacks/spica/packages/core/route/route.ts
```diff
@@ -22,6 +22,8 @@ export interface Route {
   data?: {[key: string]: any};
   queryParams?: {[key: string]: any};
   index?: number;
+  group?: string;
```
htuna07, line 25: You can rename this key as resourceCategory.

PR #862, opened 2022-07-20
feat(bucket): new ui of bucket index

akhnos:


stacks/spica/src/bucket/pages/index/index.component.html
```diff
@@ -115,21 +102,27 @@ <h5>
...
-            >
-              <mat-checkbox
+            <ng-container *ngFor="let property of properties">
+              <button
+                mat-menu-item
+                disableRipple
+                *ngIf="
+                  property.name != '$$spicainternal_select' &&
+                  property.name != '$$spicainternal_id' &&
+                  property.name != '$$spicainternal_actions'
```
htuna07, line 112: Not required but it would be better if we define an array for these values like **alwaysDisplayedProperties**

stacks/spica/src/bucket/pages/index/index.component.ts
```diff
@@ -323,7 +320,12 @@ export class IndexComponent implements OnInit, OnDestroy {
         "$$spicainternal_actions"
       ];
     } else {
-      this.displayedProperties = [schema.primary, "$$spicainternal_actions"];
+      this.displayedProperties = [
+        "$$spicainternal_select",
+        "$$spicainternal_id",
+        schema.primary,
+        "$$spicainternal_actions"
+      ];
```
htuna07, line 328: We can pass an array items if we define an array for these properties

htuna07 (commented):
There are some additional issues to be resolved before approving this pull request.
- Failed tests
- If one of the fields is in editing mode, pressing the tab key should move focus to the next field, but it seems like it's not working

htuna07 (approved):

PR #911, opened 2022-11-24
fix(storage): remove public access of the storage view endpoints

htuna07:


htuna07:
https://github.com/spica-engine/spica/pull/1149

PR #915, opened 2022-12-06
feat(bucket): full sized storage image preview on bucket-data edit page

Davuttrg:


akhnos (approved):

htuna07 (changes requested):
Previewing existing storage objects of bucket-data and previewing after drag-drop works fine, but when I try to click the edit button, select another storage object, then click to preview before clicking the save button, the "Could not load the object" error is occurring.

stacks/spica/src/storage/components/storage/storage.component.ts
```diff
@@ -83,12 +86,12 @@ export class StorageComponent implements ControlValueAccessor {
     if (!obj) {
       return;
     }
+    console.log("obj :", obj);
```
htuna07, line 89: We should remove this line before merging the PR.

htuna07 (approved):

PR #916, opened 2022-12-06
feat(bucket): display field type icons in columns on bucket-data list

Davuttrg:


stacks/spica/src/bucket/pages/index/index.component.ts
```diff
@@ -186,6 +194,9 @@ export class IndexComponent implements OnInit, OnDestroy {
       }),
       tap(schema => {
         Object.keys(schema.properties).map(key => {
+          schema.properties[key]["icon"] = this.systemFields.find(
```
akhnos, line 197: Instead of manipulating the original schema.properties, we should manipulate the deep copy object. Also, instead of using schema.properties in view, we should use a deep copy object there as well. As I see, using `displayedProperties` will be a better implementation.

akhnos (changes requested):

akhnos (approved):

htuna07 (dismissed):

stacks/spica/src/bucket/pages/index/index.component.ts
```diff
@@ -109,6 +110,7 @@ export class IndexComponent implements OnInit, OnDestroy {
   displayTranslateButton = false;
 
   private postRenderingQueue: Array<any> = [];
+  systemFields: InputPlacerWithMetaPlacer[] = [];
```
htuna07, line 113: Sorry, but I couldn't see where we use this property. If we don't use it, can you remove it, please?

stacks/spica/src/bucket/pages/index/index.component.ts
```diff
@@ -170,6 +173,11 @@ export class IndexComponent implements OnInit, OnDestroy {
           localStorage.getItem(`${this.bucketId}-displayedProperties`)
         );
 
+        this.systemFields = [];
+        this._inputResolver
+          .entries()
+          .map(e => this.systemFields.push(this._inputResolver.resolve(e)));
+
```
htuna07, line 180: Also, these lines seem unnecessary too.

stacks/spica/src/bucket/pages/index/index.component.html
```diff
@@ -187,7 +187,7 @@ <h6>
         </ng-container>
         <ng-container matColumnDef="$$spicainternal_id" *ngIf="schema$ | async; let schema" sticky>
           <th mat-header-cell *matHeaderCellDef mat-sort-header>
-            _id
+            <mat-icon class="type-icon"> key</mat-icon> _id
```
htuna07, line 190: We can remove **type-icon** class if we don't have any CSS for it.

htuna07 (changes requested):

htuna07 (approved):

PR #917, opened 2022-12-08
feat(bucket): more flexible resizing column width

Davuttrg:


akhnos (changes requested):
It is better to modify the existing resize directive instead of introducing a new directive. Please apply the required changes to the existing resize directive.

htuna07:
Also please ensure that all the current functionalities(storing the width of columns on the storage, resizing the column width should not change other column widths etc.) are still working when PR is ready

stacks/spica/packages/material/resize/resize.directive.ts
```diff
@@ -91,13 +99,33 @@ export class MatResizeHeader implements AfterViewInit {
...
+      this.resize.emit(this._width);
     }
   }
+  removeIntervalRowClass() {
+    this.renderer.removeClass(this.elemRow, this.elemRowClass);
+  }
+
+  addIntervalRowClass() {
+    this.renderer.listen("document", "mousemove", event => {
+      if (this.hasIntervalClass()) this.setCellWidth(event);
```
akhnos, line 120: Instead of if-check here, it will be better to remove the listener in removeIntervalRowClass. Otherwise, it cause triggering multiple closure functions

akhnos (changes requested):

htuna07 (approved):

akhnos (approved):

PR #919, opened 2022-12-14
feat(storage): directory support for storage objects

htuna07:
Todo

- [x] Write a migration that adds a prefix(might be 'root') to all storage objects to list them under the root directory.
- [x] Uploaded by drag&drop objects from the bucket-data edit page should have prefixes. We can use the name of the bucket for example.
- [x] Client code should run for objects even if their parents don't exist on the API. For example, if the user sends a storage object with the name "root/test" to the REST API, the client code should run even if the user forgets to send "root/" object. Especially I should remove _id check and use full name to compare, update etc.


stacks/api/storage/src/storage.service.ts
```diff
@@ -59,50 +54,62 @@ export class StorageService extends BaseCollection<StorageObject>("storage") {
...
+      this.estimatedDocumentCount()
+    )).result();
+
+    if (paginate) {
+      return this._coll
+        .aggregate<PaginatedStorageResponse>(pipeline)
+        .next()
+        .then(async r => {
+          if (!r.data.length) {
+            r.meta = {count: 0};
```
htuna07, line 95: I think I should use the **total** keyword instead of the **count**.

stacks/spica/src/storage/components/add-directory-dialog/add-directory-dialog.component.html
```diff
@@ -0,0 +1,27 @@
...
+      pattern="^[^\/\ \.]+"
+    />
+    <mat-error *ngIf="nameModel?.errors?.nameExists">
+      This name already exists.
+    </mat-error>
+    <mat-error *ngIf="nameModel?.errors?.pattern">
+      Name must not include slash, space or dot characters.
+    </mat-error>
+  </mat-form-field>
+  <button mat-button [disabled]="nameModel.invalid" (click)="dialogRef.close(name)">
```
Davuttrg, line 24: I didn't see the can-interact directive. Have you tested it without the storage policy?

htuna07: Fixed at [this](https://github.com/spica-engine/spica/pull/919/commits/7f0935d92a4f8e38c290ec6d2407cc93c486c732) commit.

stacks/spica/src/storage/components/picker/picker.component.ts
```diff
@@ -30,11 +31,13 @@ export class PickerComponent implements OnInit {
   ngOnInit(): void {
     this.storages$ = merge(this._paginator.page, of(null), this.refresh).pipe(
```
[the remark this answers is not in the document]
htuna07: Fixed at last commit.

akhnos (dismissed):

stacks/spica/src/storage/components/storage/storage.component.ts
```diff
@@ -52,16 +55,22 @@ export class StorageComponent implements ControlValueAccessor {
...
-            return Math.round((100 * event.loaded) / event.total);
-          } else if (event.type === HttpEventType.Response) {
-            this.value = event.body[0].url;
-            this.onChangeFn(this.value);
-            this.progress$ = undefined;
-          }
-        }),
+
+      this.progress$ = this.rootDir.find("root").pipe(
+        switchMap(r => (r ? of(r) : this.rootDir.add("root").toPromise())),
```
Davuttrg, line 60: No need to use "switchMap" operator cause you are not waiting for anything.
Try using it as I wrote below.

  ```
   this.progress$ = this.rootDir.find("root").pipe(
        tap((r)=>!r &&  this.rootDir.add("root").toPromise())
        switchMap(() =>
```

Davuttrg (approved):

PR #946, opened 2023-05-31
perf(function): uninstall only necessary dependencies

htuna07:
These changes will affect dependency install, uninstall on asset and versioncontrol operations only. It has no effect on install, uninstall dependency from function endpoints.


htuna07:
These changes will affect dependency install, uninstall on asset and versioncontrol operations only. It has no effect on install, uninstall dependency from function endpoints.

akhnos (approved):

PR #966, opened 2023-10-19
fix(core): update objectid validation

htuna07:


htuna07:
We decided to close this PR. This issue won't be valid after we upgrade MongoDB to the latest version.

PR #968, opened 2023-10-27
feat(status): add method and url to the request-response stats

htuna07:


htuna07:
We decided these stats should not be tracked here because of the DB performance. Instead, we should use stdout to log request stats when enabled.

PR #974, opened 2023-12-12
fix(charts): retry replicaset init command

htuna07:
If mongodb process is not ready when replicaset initiation command is executed, it throws an error like "can't connect to the server.". Instead of waiting for 5 seconds, we check whether the server is ready.


htuna07:
[This PR](https://github.com/spica-engine/spica/pull/975) provides a better solution for initiating replicaset.

PR #975, opened 2023-12-28
feat(charts): start mongodb in auth mode

htuna07:
- [x] Get rid of preparing mongodb connection uri on the helm charts, it's less secure than passing credentials as a secret key reference to the containers. Api might check environment variables of mongodb credentials and add them to the base connection string if exist. They will be injected into the container by Kubernetes secret references, so credential injection will be more secure than the current approach. The same approach should be applied to the replication controller as well. Also providing credentials via helm chart values makes them easier to read on the system. Both spica and root credentials should be provided by secret name, not plain-text.


charts/spica/templates/_helpers.tpl
```diff
@@ -1,5 +1,6 @@
 {{- define "database.connection-uri" -}}
     {{- $uri := "mongodb://" -}}
+    {{- $uri = printf "%s%s:%s@" $uri .Values.database.username .Values.database.password }}
```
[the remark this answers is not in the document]
htuna07: Instead, set the api container environment for mongodb username and password

.prettierignore
```diff
@@ -0,0 +1,2 @@
+charts/spica/**/*.yaml
+charts/spica/**/*.yml
```
mavliut-fozliev, line 2: does prettier still give an error?

mavliut-fozliev (approved):

.prettierignore
htuna07: Yes, prettier can't format yaml files under the helm charts because they include some template syntax like `{{- if}}, {{ Values...}}` etc. It's not valid for standard yaml files so that's why prettier throws an error and can't format, but it's valid for helm charts, thats why I ignored them.

PR #1047, opened 2024-12-20
test: add configuration for api/bucket

mavliut-fozliev:


apps/api/src/bucket/test/bucket.controller.spec.ts
```diff
@@ -235,12 +233,12 @@ describe("BucketController", () => {
       const secondInsertedBucket = (await req.post("/bucket", bucket2)).body;
 
       let {body: insertedDocument} = await req.post(`/bucket/${secondInsertedBucket._id}/data`, {
-        title: "title",
-        description: "description"
+        name: "name",
+        age: 18
       });
```
htuna07, line 238: We should check the reason for this change.

apps/api/src/bucket/test/bucket.controller.spec.ts
```diff
@@ -10,7 +10,7 @@ import {DatabaseTestingModule, ObjectId} from "@spica-server/database/testing";
 import {PassportTestingModule} from "@spica-server/passport/testing";
 import {PreferenceTestingModule} from "@spica-server/preference/testing";
 
-jasmine.DEFAULT_TIMEOUT_INTERVAL = 10000;
```
htuna07, line 13: We could put this configuration to the somewhere global.

htuna07 (changes requested):
Will be ready after the changes I mentioned

PR #1063, opened 2024-12-25
test: move function webhook tests

htuna07:


apps/api/src/function/webhook/test/invoker.spec.ts
```diff
@@ -185,22 +193,20 @@ describe("Webhook Invoker", () => {
       webhook: hook._id.toHexString(),
       succeed: false
     } as any);
-  });
+  }, 60_000);
```
htuna07, line 196: I will delete it

mavliut-fozliev (approved):

PR #1068, opened 2024-12-27
test: add configuration for api/function/runtime

mavliut-fozliev:


apps/api/src/function/runtime/test/node/entrypoint.spec.ts
```diff
@@ -255,35 +249,40 @@ describe("Entrypoint", () => {
...
+        .catch(e => e);
     });
-    queue.enqueue(ev);
-
-    const exitCode = await spawn().catch(e => e);
-    expect(exitCode).toBe(4);
   });
 
   it("should set env variables from scheduling context", async () => {
+    console.time("initFn");
```
htuna07, line 275: Could you remove these console time lines?

nx.json
```diff
@@ -33,5 +33,15 @@
...
+  },
+  "targetDefaults": {
+    "test": {
+      "dependsOn": [
+        {
+          "projects": ["api"],
+          "target": "build-workers"
+        }
+      ]
+    }
```
htuna07, line 45: We can remove this configuration for now and decide whether we should add it later.

htuna07 (changes requested):
Will be okay after change requests

htuna07 (approved):

PR #1076, opened 2024-12-31
feat: migrate angular to version 15

mavliut-fozliev:


apps/spica/environments/environment.ts
```diff
@@ -5,5 +5,5 @@
 
 export const environment = {
   production: true,
-  api: "http://localhost:4300"
```
htuna07, line 8: Could you revert this?

apps/spica/src/styles.scss
```diff
@@ -86,19 +98,19 @@ mat-card {
...
     }
   }
-  .mat-form-field-label {
+  .mdc-form-field__label {
     top: 2.14375em !important;
   }
-  &.mat-form-field-appearance-outline.mat-form-field-can-float.mat-form-field-should-float {
-    .mat-form-field-label {
+  &.mdc-form-field--outlined.mdc-form-field--can-float.mdc-form-field--should-float {
+    .mdc-form-field__label {
```
htuna07, line 113: Could you explain or refer to why these class names include kebab and snake case naming? Also, some of them include double dashes between words, are they running as expected?

htuna07 (commented):
Needs clarification

apps/spica/src/styles.scss
mavliut-fozliev: This is the result of the migration script. 
Screenshot 2024-12-31 at 11 06 59

htuna07 (approved):

PR #1092, opened 2025-01-09
chore: upgrade root npm packages

mavliut-fozliev:


apps/api/src/function/packages/internal_common/src/request.ts
```diff
@@ -27,7 +27,11 @@ export class Axios implements HttpService {
 
   private readonly interceptors = {
     request: {
-      onFulfilled: (request: AxiosRequestConfig) => {
+      onFulfilled: (
+        request: Omit<AxiosRequestConfig, "headers"> & {
+          headers: any; // headers type from axios 0.x.x
+        }
+      ) => {
```
htuna07, line 34: Could you explain why we need such a change?

apps/api/src/function/src/engine.ts
```diff
@@ -146,7 +146,7 @@ export class FunctionEngine implements OnModuleInit, OnModuleDestroy {
 
   deleteFunction(fn: Function) {
     const functionRoot = path.join(this.options.root, fn._id.toString());
-    return util.promisify(rimraf)(functionRoot);
+    rimraf(functionRoot);
```
htuna07, line 149: We changed the return type of this function with these changes, please confirm that none of the code that calls this function is affected.

mavliut-fozliev: fixed

htuna07 (approved):

PR #1097, opened 2025-01-14
chore: update npm package mongodb to version 4

mavliut-fozliev:


apps/api/src/function/packages/database/src/database.ts
```diff
@@ -37,13 +35,10 @@ async function connect(): Promise<_mongodb.MongoClient> {
...
     connection = new mongodb.MongoClient(process.env.__INTERNAL__SPICA__MONGOURL__, {
       replicaSet: process.env.__INTERNAL__SPICA__MONGOREPL__,
-      appname: `Functions on ${process.env.RUNTIME || "unknown"} runtime.`,
-      useNewUrlParser: true,
-      // @ts-ignore
-      useUnifiedTopology: true
+      appName: `Functions on ${process.env.RUNTIME || "unknown"} runtime.`
     });
   }
-  if (!connection.isConnected()) {
```
htuna07, line 46: Could you add a link that refers isConnected method removed and all connections are considered as connected?

apps/api/src/preference/src/preference.controller.ts
```diff
@@ -19,6 +19,7 @@ import {
   IDENTITY_SETTINGS_FINALIZER
 } from "@spica-server/preference/services";
 import {createPreferenceActivity} from "./activity.resource";
+import {ReturnDocument} from "mongodb";
```
htuna07, line 22: I think we should import it from spica database package.

libs/database/collection.ts
```diff
@@ -64,7 +68,7 @@ export class _MixinCollection<T> {
 
   aggregate<ResponseType>(
     pipeline?: object[],
-    options: CollectionAggregationOptions = {allowDiskUse: true}
```
htuna07, line 67: I think the options could be more than allow disk use, is there any mongodb interface that we can use for aggregate options?

libs/database/collection.ts
```diff
@@ -80,80 +84,87 @@ export class _MixinCollection<T> {
...
+    const result: InsertOneResult<T> = await this._coll.insertOne(doc);
+    doc._id = result.insertedId;
+    return doc as WithId<T>;
   }
 
-  async insertMany(docs: Array<T>): Promise<ObjectId[]> {
+  async insertMany(docs: Array<OptionalUnlessRequiredId<T>>): Promise<ObjectId[]> {
     await this.documentCountLimitValidation(docs.length);
 
-    return this._coll.insertMany(docs).then(t => Object.values(t.insertedIds));
```
htuna07, line 92: Could you explain why we made such a change? Seems they are identical.

libs/database/collection.ts
```diff
@@ -80,80 +84,87 @@ export class _MixinCollection<T> {
...
-  findOneAndDelete(filter: FilterQuery<T>, options?: FindOneAndDeleteOption): Promise<T> {
-    return this._coll.findOneAndDelete(filter, options).then(r => r.value);
+  findOneAndDelete(filter: Filter<T>, options?: FindOneAndDeleteOptions): Promise<WithId<T>> {
+    return this._coll.findOneAndDelete(filter, options);
   }
 
-  deleteOne(filter: FilterQuery<T>, options?: FindOneAndDeleteOption): Promise<number> {
-    return this._coll.deleteOne(filter, options).then(r => r.deletedCount);
+  deleteOne(filter: Filter<T>, options?: FindOptions): Promise<number> {
+    return this._coll.deleteOne(filter, options).then(r => r.deletedCount || 0);
```
htuna07, line 117: Is it possible that deletedCount is undefined?

libs/database/collection.ts
```diff
@@ -1,26 +1,30 @@
...
 export interface InitializeOptions {
   entryLimit?: number;
-  collectionCreateOptions?: CollectionCreateOptions;
-  afterInit?: (...args) => any;
+  collectionOptions?: CollectionOptions;
+  afterInit?: (...args: any[]) => any;
 }
 
-export type OptionalId<T> = Omit<T, "_id"> & {_id?: ObjectId | string | number};
+export type _OptionalId<T> = Omit<T, "_id"> & {_id?: ObjectId | string | number};
```
htuna07, line 27: Why did we make such a change?

libs/database/testing/start.ts
```diff
@@ -38,10 +38,7 @@ export function getDatabaseName() {
 }
 
 function getClientOptions(): MongoClientOptions {
```
htuna07, line 40: We could remove this method instead of returning nothing.

libs/database/database.module.ts
```diff
@@ -15,7 +15,7 @@ export class DatabaseModule {
         useFactory: async () => {
           const opts = {...options};
           delete opts.database;
-          return MongoClient.connect(uri, {...opts, useNewUrlParser: true});
+          return MongoClient.connect(uri, {...opts});
```
htuna07, line 18: Why not just 
**return MongoClient.connect(uri, opts);**

htuna07 (changes requested):
Waiting for change requests and answers.

apps/api/src/function/packages/database/src/database.ts
mavliut-fozliev: https://github.com/mongodb/node-mongodb-native/blob/13ca440500b515e9aaf0aef4409503c518d13413/etc/notes/CHANGES_4.0.0.md#removed-deprecations

htuna07: Okay, we should also update this file based on the removed methods then. For example, we can remove the if check before calling the connect method of the connection, since calling connect for already connected connections has no effect(what the documentation says). Feel free to make these kinds of updates for redundant parts.

htuna07 (changes requested):

htuna07 (approved):

PR #1100, opened 2025-01-16
chore: upgrade npm package mongodb to version 6

mavliut-fozliev:


apps/api/src/preference/services/preference.service.ts
```diff
@@ -62,16 +62,13 @@ export class PreferenceService extends BaseCollection("preferences") {
     preference: T,
     options?: FindOneAndReplaceOptions
   ) {
-    return this._coll
-      .findOneAndReplace(filter, preference, options)
-      .then(preference => preference.value);
+    return this._coll.findOneAndReplace(filter, preference, options).then(preference => preference);
```
htuna07, line 65: The **then** method seems redundant.

htuna07 (changes requested):

apps/api/src/preference/services/preference.service.ts
mavliut-fozliev: removed

htuna07 (approved):

PR #1113, opened 2025-01-21
build(deps): upgrade api npm packages for big major changes

mavliut-fozliev:


apps/api/src/passport/guard/src/action.guard.ts
```diff
@@ -88,13 +88,18 @@ function buildResourceAndModuleName(path: string, params: object, format?: strin
...
     path = format;
   }
-  const segments = parse(path);
+  const segments = parse(path).tokens;
+
   const resourceSegments = segments
-    .filter(s => typeof s == "string")
-    .map((s: string) => s.replace(/^\//g, ""))
+    .filter(s => s.type == "text")
+    .map((s: Text) => s.value.replace(/\//g, ""))
```
htuna07, line 95: Why did we change the first replace argument?

apps/api/package.json
```diff
@@ -14,14 +14,13 @@
     "archiver": "^5.3.1",
     "aws-sdk": "^2.1206.0",
     "bcryptjs": "^2.4.3",
-    "cache-manager": "^3.4.1",
```
htuna07, line 17: Could you explain why we don't need it anymore if it's an intended change?

htuna07 (approved):

apps/api/package.json
mavliut-fozliev: I checked the source code and found that we do not use this package, we only use @types/cache-manager which is a separate npm package.

apps/api/src/passport/guard/src/action.guard.ts
mavliut-fozliev: in the old version of path-to-regexp tokens have slashes at the beginning of the line, whereas in the new version of path-to-regexp tokens have slashes on both sides.

PR #1138, opened 2025-02-03
build: add build and publish configurations for devkit packages and cli

mavliut-fozliev:


apps/api/src/function/packages/bucket/project.json
```diff
@@ -1,5 +1,20 @@
 {
   "name": "api/function/packages/bucket",
   "sourceRoot": "apps/api/src/function/packages/bucket",
-  "projectType": "library"
+  "projectType": "library",
```
htuna07, line 4: Please execute build first, then resolve ts paths.

mavliut-fozliev: done

htuna07 (approved):

PR #1140, opened 2025-02-04
build(image): add docker image configuration for "migrate" project

mavliut-fozliev:


apps/migrate/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+RUN yarn nx build migrate
+
+FROM base AS test
+WORKDIR /app
+COPY --from=deps /app/node_modules ./node_modules
+COPY apps/migrate ./apps/migrate
+COPY nx.json ./
+COPY package.json ./
+COPY tsconfig.json ./
+COPY apps/migrate/tsconfig.spec.json ./
```
htuna07, line 34: Shouldn't line 30 copy-paste entire apps/migrate directory? Confirm whether this line is really necessary plase.

apps/migrate/project.json
```diff
@@ -63,6 +63,33 @@
       },
       "dependsOn": ["build-test"],
       "cache": true
+    },
+    "test-image": {
```
htuna07, line 67: You don't need to tag the docker test output.

apps/migrate/tsconfig.spec.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "./tsconfig.json",
```
htuna07, line 2: Why did we need such a change?

htuna07 (changes requested):

apps/migrate/Dockerfile
mavliut-fozliev: fixed

apps/migrate/project.json
mavliut-fozliev: fixed

htuna07 (approved):

PR #1184, opened 2025-02-19
chore(migrate): add migration script for storage objects with the same name

mavliut-fozliev:


apps/migrate/src/migrations/0.9.30/add_index_for_duplicate_storage_objects.ts
```diff
@@ -0,0 +1,36 @@
+import {Context} from "../../migrate";
+
+export default async function (ctx: Context) {
+  const coll = ctx.database.collection("storage");
+  const storageObjects = await coll.find().toArray();
```
htuna07, line 5: Finding only storage objects with the same names will consume less resources than fetching the entire collection.

htuna07 (changes requested):
Waiting for changes

apps/migrate/src/migrations/0.9.30/add_index_for_duplicate_storage_objects.ts
mavliut-fozliev: fixed

htuna07 (approved):

htuna07 (approved):

PR #1188, opened 2025-02-21
build(api): use slim version of node

mavliut-fozliev:
I try to use node:22-alpine to reduce image size, but there are errors in tests because mongodb-memory-server package does not work on this version of node image.


mev-isa:
I use node:22-slim with the addition of the git package for the build and libcurl4 for running mongo-memory-server, the image size became about 175 MB.

htuna07 (approved):

PR #1190, opened 2025-02-24
fix(api): import function entrypoint with correct extension

mavliut-fozliev:


htuna07:
Because of the performance impacts of this approach, also importing another function feature is pending, we shouldn't complete this task.

PR #1191, opened 2025-02-24
refactor(api): drop using deprecated method for removing directories

mavliut-fozliev:


htuna07:
Could you check the link below to complete the task fully?
https://nodejs.org/api/deprecations.html#DEP0147

htuna07 (approved):

PR #1200, opened 2025-02-26
feat(api): introduce json field type

mavliut-fozliev:


htuna07:
Please add tests for this feature. Tests should include GET, POST, PUT and PATCH requests, also I wonder whether filtering for this field will work.

htuna07 (changes requested):
Waiting for changes

apps/api/src/bucket/test/bucket-data.controller.spec.ts
```diff
@@ -1988,4 +1988,117 @@ describe("BucketDataController", () => {
...
+    it("should update document", async () => {
+      await req.patch(`/bucket/${bucket._id}/data/${rows[0]._id}`, {
+        characteristics: {height: 200}
+      });
+      const document = await req.get(`/bucket/${bucket._id}/data/${rows[0]._id}`);
+
+      const {statusCode, statusText, body} = document;
+
+      expect([statusCode, statusText]).toEqual([200, "OK"]);
+      expect(body).toEqual({...body, characteristics: {...body.characteristics, height: 200}});
```
htuna07, line 2064: Using the response body for the comparison of both expected and actual sides might silence some unexpected behaviors. For example, if your update request updates some other part than you sent, for example, the name property, you can't detect this issue with this comparison. Instead, prefer writing the right side of the comparison explicitly, or using rows[0].

apps/api/src/bucket/test/bucket-data.controller.spec.ts
```diff
@@ -1988,4 +1988,117 @@ describe("BucketDataController", () => {
...
+      };
+
+      const res = await req.post(`/bucket/${bucket._id}/data`, newDocument);
+      expect([res.statusCode, res.statusText]).toEqual([201, "Created"]);
+
+      const document = await req.get(`/bucket/${bucket._id}/data/${res.body._id}`);
+      const {statusCode, statusText, body} = document;
+
+      expect([statusCode, statusText]).toEqual([200, "OK"]);
+      expect(body).toEqual({...newDocument, _id: body._id});
```
htuna07, line 2052: Why don't you use res.body._id for the toEqual part to ensure post response id and id of the data returned from get request are the same?

apps/api/src/bucket/test/bucket-data.controller.spec.ts
```diff
@@ -1988,4 +1988,117 @@ describe("BucketDataController", () => {
...
+        characteristics: {height: 200, weight: "80kg"}
+      };
+
+      await req.put(`/bucket/${bucket._id}/data/${rows[0]._id}`, newDocument);
+      const document = await req.get(`/bucket/${bucket._id}/data/${rows[0]._id}`);
+
+      const {statusCode, statusText, body} = document;
+
+      expect([statusCode, statusText]).toEqual([200, "OK"]);
+      expect(body).toEqual({...newDocument, _id: body._id});
```
htuna07, line 2079: Use rows[0]._id instead of body._id for comparison.

htuna07 (changes requested):
Waiting for changes

apps/api/src/bucket/test/bucket-data.controller.spec.ts
mavliut-fozliev: corrected

apps/api/src/bucket/test/bucket-data.controller.spec.ts
mavliut-fozliev: corrected

apps/api/src/bucket/test/bucket-data.controller.spec.ts
```diff
@@ -1988,4 +1988,120 @@ describe("BucketDataController", () => {
...
+      const {statusCode, statusText, body} = document;
+
+      expect([statusCode, statusText]).toEqual([200, "OK"]);
+      expect(body).toEqual({
+        ...rows[0],
+        characteristics: {...rows[0].characteristics, height: 200}
+      });
+    });
+
+    fit("should replace document", async () => {
```
htuna07, line 2070: Could you remove the fit?

htuna07 (changes requested):
Waiting for changes

htuna07 (approved):

apps/api/src/bucket/test/bucket-data.controller.spec.ts
mavliut-fozliev: removed

htuna07 (approved):

PR #1211, opened 2025-03-04
feat(api): implement oauth strategy named custom

mavliut-fozliev:


apps/migrate/src/migrations/0.13.1/add_idp_for_oauth_strategies.ts
```diff
@@ -0,0 +1,18 @@
...
+  for (let strategy of strategies) {
+    bulkOps.push({
+      updateOne: {filter: {_id: strategy._id}, update: {$set: {"options.idp": "custom"}}}
+    });
+  }
+
+  if (bulkOps.length) {
+    await coll.bulkWrite(bulkOps);
+  }
+}
```
htuna07, line 18: Could you use one call to update only the oauth strategies idps?

htuna07 (changes requested):
Waiting for changes

apps/migrate/src/migrations/0.13.1/add_idp_for_oauth_strategies.ts
mavliut-fozliev: corrected

htuna07 (approved):

PR #1212, opened 2025-03-04
build(api): rename typescript function build output

mavliut-fozliev:


apps/api/src/function/compiler/typescript/src/typescript_worker.ts
```diff
@@ -128,7 +128,21 @@ function build(compilation: Compilation) {
...
 
+function renameJsToMjs(baseUrl: string) {
+  const buildFolder = path.join(baseUrl, ".build");
+  const filePath = path.join(buildFolder, "index.mjs");
+
+  fs.renameSync(path.join(buildFolder, "index.js"), filePath);
+  fs.renameSync(path.join(buildFolder, "index.js.map"), path.join(buildFolder, "index.mjs.map"));
+
+  let content = fs.readFileSync(filePath, "utf8");
+  content = content.replace(/(\/\/# sourceMappingURL=).*\.map$/, "$1index.mjs.map");
```
htuna07, line 139: could you check that mjs map renaming really necessary?

htuna07 (changes requested):
Waiting for changes

apps/api/src/function/compiler/typescript/src/typescript_worker.ts
mavliut-fozliev: mjs map renaming is not necessary

htuna07 (approved):

PR #1216, opened 2025-03-05
feat(api): introduce oauth identity providers presets

htuna07:


htuna07:
Wait until all presets are introduced.

PR #1232, opened 2025-03-24
chore(migrate): add migration script for function env vars

mavliut-fozliev:


htuna07 (changes requested):
We don't store environment variables in the function schema like this, please use the correct form.

htuna07 (approved):

PR #1234, opened 2025-03-24
feat: introduce asset management for environment variables

mavliut-fozliev:


apps/api/src/env_var/test/controller.spec.ts
```diff
@@ -208,18 +208,5 @@ describe("Environment Variable", () => {
...
-        key: "ENV_KEY",
-        value: "123"
-      });
-
-      const res = await req.delete(`/env-var/${new ObjectId()}`);
-      expect([res.body.statusCode, res.body.message]).toEqual([404, "Not Found"]);
-
-      const {body} = await req.get("/env-var");
-      expect(body.length).toEqual(1);
-    });
```
htuna07, line 223: Why did we remove this test?

mavliut-fozliev: I saw that when deleting data from a bucket with a non-existent id, the error does not return and did the same.

htuna07: Please add a decorator to return the "No Content" then, just like we do for bucket.

htuna07 (changes requested):
Waiting for changes

htuna07 (approved):

PR #1240, opened 2025-03-26
refactor: move api interfaces, types and tokens to libs

mavliut-fozliev:


htuna07 (commented):
Will be approved after conflicts are resolved

htuna07 (approved):

PR #1244, opened 2025-03-27
feat: resolve relations of realtime changes

mavliut-fozliev:


apps/api/src/function/packages/bucket/src/bucket.ts
```diff
@@ -146,21 +146,39 @@ export namespace data {
   }
 
   export namespace realtime {
+    /**
+     * @param {Object} queryParams - Query params.
+     * * @param {string} queryParams.relation - This will increase response time.
```
htuna07, line 151: If the relation could be boolean or string array, why did we use only string here?

htuna07: Use the description below instead,
"Resolve relations of document(s), but it will increase response time."

apps/api/src/function/packages/bucket/src/operators.ts
```diff
@@ -132,7 +139,18 @@ export function getWsObs<T>(
       return of(null);
     }),
     debounceTime(1),
-    map(() => (targetDocumentId ? Array.from(data)[0] : Array.from(data)))
+    map(() => (targetDocumentId ? Array.from(data)[0] : Array.from(data))),
+    map(chunk => {
```
htuna07, line 143: Using the switchmap operator would be more appropriate.

apps/api/src/function/packages/bucket/test/bucket.integration.spec.ts
```diff
@@ -415,6 +415,129 @@ describe("Bucket", () => {
...
+          const subject = Bucket.data.realtime.getAll(usersBucketId, {relation: true});
+
+          subject.pipe(bufferCount(2), take(1)).subscribe({
+            next: messages => {
+              Promise.all(messages).then(data => {
+                expect(data).toEqual([
+                  // initial docs
+                  [],
+                  // docs after changes
+                  [{_id: insertedUserId, name: "Joe", age: 34, headphone}]
```
htuna07, line 484: Could you write the headphone object more explicitly, like the user?

apps/api/src/function/packages/bucket/test/bucket.integration.spec.ts
```diff
@@ -415,6 +415,129 @@ describe("Bucket", () => {
...
+              relation: true
+            });
+            subject.pipe(bufferCount(2), take(1)).subscribe({
+              next: messages => {
+                Promise.all(messages).then(data => {
+                  expect(data).toEqual([
+                    // initial docs
+                    {_id: insertedUserId, name: "Joe", age: 34},
+                    // docs after changes
+                    {_id: insertedUserId, name: "Joe", age: 34, headphone}
```
htuna07, line 520: Could you write the headphone object more explicitly, like the user?

htuna07 (changes requested):
Will be okay after changes.

apps/api/src/function/packages/bucket/src/bucket.ts
mavliut-fozliev: fixed in commit

apps/api/src/function/packages/bucket/src/operators.ts
mavliut-fozliev: fixed in commit

apps/api/src/function/packages/bucket/test/bucket.integration.spec.ts
mavliut-fozliev: fixed in commit

apps/api/src/function/packages/bucket/test/bucket.integration.spec.ts
mavliut-fozliev: fixed in commit

htuna07 (approved):
LGTM

PR #1255, opened 2025-04-14
feat(api): add activity tracking to function index, dependency and environment variable endpoints

mavliut-fozliev:


htuna07 (commented):
Could you add tests?

htuna07 (approved):

htuna07 (dismissed):

PR #1256, opened 2025-04-14
refactor: add missing methods to crud.ts files

mavliut-fozliev:


apps/api/src/bucket/src/crud.ts
```diff
@@ -6,6 +6,14 @@ import {HistoryService} from "@spica-server/bucket/history";
 import * as expression from "@spica-server/bucket/expression";
 import {BadRequestException} from "@nestjs/common";
 
+export async function find(bs: BucketService, resourceFilter: object): Promise<Bucket[]> {
+  return bs.aggregate<Bucket>([resourceFilter, {$sort: {order: 1}}]).toArray();
```
htuna07, line 10: Instead of setting the sort here, we should pass it as an argument like resourceFilter. You can pass an object named **options**, which includes all arguments like resourceFilter, sort etc.

apps/api/src/function/src/function.controller.ts
```diff
@@ -221,17 +221,7 @@ export class FunctionController {
...
-      throw new NotFoundException("Can not find function.");
-    }
-    const index = await this.engine.read(fn).catch(e => {
-      if (e == "Not Found") {
-        throw new NotFoundException("Index does not exist.");
-      }
-      throw new InternalServerErrorException(e);
-    });
-    return {index};
+    return CRUD.index.find(this.fs, this.engine, id);
```
htuna07, line 224: If this line throws an error, we should catch it with something like below
`throw new HttpException(error.message, error.status || 500);`

Otherwise api will be down. Could you test it?

apps/api/src/function/src/function.controller.ts
```diff
@@ -241,11 +231,7 @@ export class FunctionController {
   @Get(":id/dependencies")
   @UseGuards(AuthGuard(), ActionGuard("function:show", "function/:id"))
   async getDependencies(@Param("id", OBJECT_ID) id: ObjectId) {
-    const fn = await this.fs.findOne({_id: id});
-    if (!fn) {
-      throw new NotFoundException("Could not find the function.");
-    }
-    return this.engine.getPackages(fn);
+    return CRUD.dependencies.findOne(this.fs, this.engine, id);
```
htuna07, line 234: Same as the comment above.

htuna07 (changes requested):
Seems okay but check comments to complete PR.
Also, following parts should use the crud.ts
- Function dependency install and uninstall

apps/api/src/bucket/src/crud.ts
mavliut-fozliev: Done

apps/api/src/function/src/function.controller.ts
mavliut-fozliev: Done

apps/api/src/function/src/function.controller.ts
mavliut-fozliev: Done

htuna07 (approved):

PR #1259, opened 2025-04-16
chore(chart): configure readiness probes for pods

mavliut-fozliev:


htuna07 (changes requested):
Seems nice but values below would be better for both containers
- initialDelaySeconds: 10
- periodSeconds: 5
- timeoutSeconds: 5
- failureThreshold: 6

charts/spica/templates/application.yaml
```diff
@@ -64,6 +64,16 @@ spec:
...
+          readinessProbe:
+            httpGet:
+              path: /passport/strategies
+              port: 4300
+              scheme: HTTP
+            initialDelaySeconds: 10
+            periodSeconds: 5
+            timeoutSeconds: 2
+            failureThreshold: 3
+            successThreshold: 1
```
htuna07, line 76: Is it necessary? If so, we could use it for the database too.

mavliut-fozliev: it is not necessary

htuna07 (approved):

PR #1260, opened 2025-04-16
chore(chart): set replicas count

mavliut-fozliev:


charts/spica/templates/application.yaml
```diff
@@ -78,7 +78,7 @@ spec:
             --passport-password,  {{ .Values.application.defaultPassword }},
             --public-url, {{ required "A valid .Values.application.publicHost is required!" .Values.application.publicHost }}{{.Values.application.apiBaseUrl | trimSuffix "/"}},
             --persistent-path, /data,
-            --replication, {{ ternary "true" "false" (gt (int .Values.application.replicas) 1) }}
+            --replication, "{{ ternary "true" "false" (gt (int .Values.application.replicas) 1) }}"
```
htuna07, line 81:
```suggestion
            --replication, "{{ ternary true false (gt (int .Values.application.replicas) 1) }}"
```

htuna07 (changes requested):
Waiting for change request

htuna07 (approved):

PR #1262, opened 2025-04-18
test: add missing tests for versioncontrol

mavliut-fozliev:


htuna07 (approved):

htuna07 (commented):
Suggestion, you can move implementations to the methods to keep tests simple and clean. For example:
`await req.post("/versioncontrol/commands/remote", {args: ["add", "origin", bareRepo]});` 
could be:
`addRemoteOrigin(bareRepo)`
instead.

PR #1268, opened 2025-04-29
feat: add RabbitMQ proto file

mavliut-fozliev:


apps/api/src/function/queue/proto/src/rabbitmq.proto
```diff
@@ -0,0 +1,28 @@
+syntax = "proto3";
+
+package RabbitMQ;
+
+message RabbitMQMessage {
```
htuna07, line 5: Please use only the Message name since it belongs to the RabbitMQ already.

mavliut-fozliev: This creates a naming conflict in the entrypoint.js file.

apps/api/src/function/queue/proto/src/rabbitmq.proto
```diff
@@ -0,0 +1,28 @@
...
+    string properties = 4;
+
+    message Pop {
+        string id = 1;
+    }
+
+    message Result {}
+}
+
+message Error {
```
htuna07, line 18: You could remove the Error message and queue call because error flow will be decided later.

htuna07 (changes requested):
Mostly seems fine

apps/api/src/function/queue/proto/src/rabbitmq.proto
mavliut-fozliev: Done

htuna07 (approved):

PR #1269, opened 2025-04-29
feat: add RabbitMQ enqueuer

mavliut-fozliev:


apps/api/src/function/enqueuer/src/rabbitmq.ts
```diff
@@ -0,0 +1,119 @@
...
+        connection
+          .createChannel()
+          .then(channel => {
+            channel.on("error", err => {
+              console.error("Channel error:", err.message);
+            });
+
+            if (options.exchange) {
+              channel.assertExchange(options.exchange.name, options.exchange.type, {
+                durable: options.exchange.durable ?? false
```
htuna07, line 41: Since it's defined on the schema, and it's boolean, you can use the provided durable value directly.

apps/api/src/function/enqueuer/src/rabbitmq.ts
```diff
@@ -0,0 +1,119 @@
...
+            }
+
+            channel
+              .assertQueue(options.queue.name, {
+                durable: options.queue.durable,
+                exclusive: options.queue.exclusive
+              })
+              .then(q => {
+                if (options.exchange) {
+                  const pattern = options.exchange.pattern ?? "";
```
htuna07, line 52: Could you define a default value on the schema and remove this line instead?

apps/api/src/function/enqueuer/src/rabbitmq.ts
```diff
@@ -0,0 +1,119 @@
...
+
+            if (options.prefetch) {
+              channel.prefetch(options.prefetch);
+            }
+
+            channel.consume(
+              options.queue.name,
+              (msg: amqp.Message | null) => this.onMessageHandler(target, msg, channel, options),
+              {
+                noAck: options.noAck ?? false
```
htuna07, line 65: Same here.

apps/api/src/function/enqueuer/src/rabbitmq.ts
```diff
@@ -0,0 +1,119 @@
...
+    this.schedulerUnsubscription(target.id);
+
+    for (const connection of this.connections) {
+      if (
+        (!target.handler && connection["target"].cwd == target.cwd) ||
+        (target.handler &&
+          connection["target"].cwd == target.cwd &&
+          connection["target"].handler == target.handler)
+      ) {
+        connection.close();
```
htuna07, line 88: Could you check that closing the connection is sufficient to clean resources? Otherwise, there might be a leak here.

apps/api/src/function/enqueuer/src/rabbitmq.ts
mavliut-fozliev: Done

apps/api/src/function/enqueuer/src/rabbitmq.ts
mavliut-fozliev: Done

apps/api/src/function/enqueuer/src/rabbitmq.ts
mavliut-fozliev: Done

apps/api/src/function/enqueuer/test/rabbitmq.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+  });
+
+  it("should subscribe", async () => {
+    rabbitmqEnqueuer.subscribe(noopTarget, {
+      url,
+      queue: {name: "queue1", durable: true},
+      noAck: true
+    });
+
+    await new Promise(r => setTimeout(r, 100));
```
htuna07, line 49: Could you increase the settimeout, since they might fail on slower computers?

apps/api/src/function/enqueuer/test/rabbitmq.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+        channel.sendToQueue(queue, Buffer.from(msg), {
+          persistent: true
+        });
+      });
+
+      setTimeout(() => {
+        connection.close();
+        process.exit(0);
+      }, 500);
+    });
```
htuna07, line 121: Could you use the promise approach instead of callbacks? Also, you could use afterEach to close the connection instead of the setTimeout, and why did we need to use process.exit?

apps/api/src/function/enqueuer/src/rabbitmq.ts
mavliut-fozliev: Done

apps/api/src/function/enqueuer/test/rabbitmq.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    await new Promise(r => setTimeout(r, 100));
+
+    expect(eventQueue.enqueue).toHaveBeenCalledTimes(1);
+    expect(rabbitmqQueue.enqueue).toHaveBeenCalledTimes(1);
+
+    const message =
+      rabbitmqQueue.enqueue.mock.calls[rabbitmqQueue.enqueue.mock.calls.length - 1][1];
+    expect(message.msg).toBe("Hello World!");
```
htuna07, line 130: Can we increase the arguments we tested, if it's possible? For example, fields and properties of the message object with channel and options arguments too.

htuna07 (changes requested):
Waiting for changes.

PR #1270, opened 2025-04-29
feat: add RabbitMQ queue

mavliut-fozliev:


apps/api/src/function/queue/src/rabbitmq.ts
```diff
@@ -0,0 +1,110 @@
...
+      if (call.request.id) {
+        const msg = {
+          content: Buffer.from(call.request.content),
+          fields: JSON.parse(call.request.fields),
+          properties: JSON.parse(call.request.properties)
+        };
+
+        serverResponse.channel.ack(msg);
+      }
+      this.dequeue(call.request.id);
```
htuna07, line 78: Could you remove this call to keep the channel for later usage?

apps/api/src/function/queue/src/rabbitmq.ts
```diff
@@ -0,0 +1,110 @@
...
+      if (call.request.id) {
+        const msg = {
+          content: Buffer.from(call.request.content),
+          fields: JSON.parse(call.request.fields),
+          properties: JSON.parse(call.request.properties)
+        };
+
+        serverResponse.channel.nack(msg);
+      }
+      this.dequeue(call.request.id);
```
htuna07, line 99: Same here

mavliut-fozliev: Done

mavliut-fozliev: Done

htuna07 (approved):

PR #1272, opened 2025-04-29
feat: create schema for trigger and add RabbitMQ support to scheduler and engine

mavliut-fozliev:


apps/api/src/function/src/schema/rabbitmq.json
```diff
@@ -0,0 +1,73 @@
+{
```
htuna07, line 1: Could you write a description for the fields, also, examples and enums would be better if possible.

mavliut-fozliev: Done

htuna07 (approved):

PR #1275, opened 2025-05-02
feat(api): introduce Rabbitmq trigger to functions

mev-isa:


apps/api/src/status/test/controller.spec.ts
```diff
@@ -283,8 +283,8 @@ describe("Status", () => {
             unit: "count"
           },
           workers: {
-            activated: 1,
-            fresh: 1,
+            activated: 0,
+            fresh: 0,
```
htuna07, line 287: We should revert this expection changes and make the test pass.

htuna07 (approved):

PR #1302, opened 2025-05-26
fix(api): remove quotes on commit message of version control module

htuna07:


apps/api/src/versioncontrol/test/e2e.spec.ts
```diff
@@ -266,10 +266,12 @@ describe("Versioning e2e", () => {
         await insertBucket(getEmptyBucket());
 
         await req.post("/versioncontrol/commands/add", {args: ["."]});
-        await req.post("/versioncontrol/commands/commit", {args: ["-m", "'first commit'"]});
+        const r = await req.post("/versioncontrol/commands/commit", {
```
mavliut-fozliev, line 269: Why do we store the result in a variable?

mavliut-fozliev (approved):
LGTM

apps/api/src/versioncontrol/test/e2e.spec.ts
htuna07: I used it for debugging and forgot to remove it, thanks.

mavliut-fozliev (approved):

PR #1312, opened 2025-05-27
fix(api): add status for newly spawned workers

mev-isa:


apps/api/src/status/test/controller.spec.ts
```diff
@@ -284,7 +284,7 @@ describe("Status", () => {
           },
           workers: {
             activated: 1,
-            fresh: 1,
+            fresh: 0,
```
htuna07, line 287: Instead of changing the behaviour, could you try to wait for one second to let api get ready workers?
Let me know if it doesn't work.

htuna07 (changes requested):
Waiting for changes

htuna07 (approved):

PR #1316, opened 2025-05-28
feat(api): refresh identity tokens

htuna07:


apps/api/src/passport/identity/src/identity.service.ts
```diff
@@ -57,10 +57,92 @@ export class IdentityService extends BaseCollection<Identity>("identity") {
...
+
+    if (refreshTokenData.identity !== String(identity._id)) {
+      return Promise.reject("Refresh and access token identifiers are mismatched");
+    }
+
+    return Promise.resolve();
+  }
+
+  private findIdentityOfToken(token: string) {
+    const {identifier} = this.decode(token);
```
mev-isa: If decode will return null we will get destructuring error

apps/api/src/passport/refresh_token/services/src/module.ts
```diff
@@ -0,0 +1,13 @@
+import {Module} from "@nestjs/common";
+import {RefreshTokenService} from "./service";
+
+@Module({})
```
mev-isa: Why we made this module dynamic?

apps/api/src/passport/refresh_token/jest.config.ts
```diff
@@ -0,0 +1,9 @@
+const path = require("path");
+const {workspaceRoot} = require("@nx/devkit");
+
+export default {
+  preset: "../../../../../jest.preset.js",
+  testEnvironment: "node",
+  coverageDirectory: path.join(workspaceRoot, "coverage/apps/api/env_var"),
```
mev-isa: Should be "coverage/apps/api/passport/refresh_token"

libs/interface/passport/refresh_token/project.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "interface/passport/refresh_token",
+  "sourceRoot": "libs/interface/passport/policy/refresh_token",
```
mev-isa: sourceRoot should be "libs/interface/passport/refresh_token/src"

htuna07: You're right, but do you know why libraries' source routes end with/src while applications don't?

apps/api/src/passport/identity/src/identity.service.ts
htuna07: I added null check on the last commit

apps/api/src/passport/refresh_token/services/src/module.ts
htuna07: It was for following the convention but it seems useless, I've made it static on the last commit.

apps/api/src/passport/refresh_token/jest.config.ts
htuna07: Done

libs/interface/passport/refresh_token/project.json
mev-isa: I think we forgot to add "src" at the end of application's sourceRoot

mavliut-fozliev (dismissed):

mavliut-fozliev (dismissed):

mavliut-fozliev (approved):

PR #1321, opened 2025-05-29
fix(api): turn apply to promise to catch error

mev-isa:


htuna07 (approved):

libs/interface/versioncontrol/src/synchronizer.ts
```diff
@@ -154,11 +154,9 @@ export abstract class Synchronizer<R1 extends Resource, R2 extends Resource> {
 
         const convertedChange = repSync.converter.convert(change);
 
-        const apply = () => repSync.applier.apply(convertedChange);
-
-        const retry = (delays: number[]) => {
+        const retry = async (delays: number[]) => {
           try {
-            apply();
+            await repSync.applier.apply(convertedChange);
```
htuna07, line 159: Could you also add more details to error like "Failed to apply changes from doc to rep for the following change...<change<...reason:<reason>"?

mev-isa: Done

htuna07 (approved):

PR #1328, opened 2025-06-03
feat(api): reduce login attempts

mavliut-fozliev:


apps/api/src/passport/test/e2e.spec.ts
```diff
@@ -895,6 +899,60 @@ describe("E2E Tests", () => {
...
+
+      // WAIT UNTIL IDENTITY IS INSERTED
+      await new Promise((resolve, _) => setTimeout(resolve, 3000));
+    });
+
+    const loginWithWrongPassword = () =>
+      req.post("/passport/identify", {
+        identifier: "spica",
+        password: "wrongPassword"
+      });
```
htuna07: You could call the login method with the wrong password instead of writing a new method.

apps/api/src/passport/test/e2e.spec.ts
```diff
@@ -895,6 +899,60 @@ describe("E2E Tests", () => {
...
+      for (const fn of Array(3).fill(loginWithWrongPassword)) {
+        responses.push(await fn());
+      }
+
+      const responseWithBlockedError = responses.at(-1);
+
+      expect(responseWithBlockedError.statusCode).toEqual(401);
+      expect(responseWithBlockedError.statusText).toEqual("Unauthorized");
+      expect(responseWithBlockedError.body.message).toContain(
+        "Too many failed login attempts. Try again after"
```
htuna07: Could you use fake timers and update this expectation to see how many seconds are left?

apps/api/src/passport/test/e2e.spec.ts
```diff
@@ -895,6 +899,60 @@ describe("E2E Tests", () => {
...
+      );
+
+      await login();
```
htuna07, line 939: You can use advanceTimersByTime to forward time, so we can ensure that the remaining seconds are decreasing as expected.

apps/api/src/passport/test/e2e.spec.ts
```diff
@@ -895,6 +899,60 @@ describe("E2E Tests", () => {
...
+    const loginWithWrongPassword = () =>
+      req.post("/passport/identify", {
+        identifier: "spica",
+        password: "wrongPassword"
+      });
+
+    it("should block if the attempt limit is exceeded ", async () => {
+      const responses = [];
+      for (const fn of Array(3).fill(loginWithWrongPassword)) {
+        responses.push(await fn());
```
htuna07, line 928: You could use the Promise.all to speed up tests.

apps/api/src/passport/test/e2e.spec.ts
```diff
@@ -895,6 +899,60 @@ describe("E2E Tests", () => {
...
+      expect(token).toBeUndefined();
+
+      jest.useFakeTimers({doNotFake: ["nextTick"]});
+      jest.advanceTimersByTime(11 * 60 * 1000);
+
+      await login();
+      jest.useRealTimers();
+      expect(token).toBeDefined();
+    });
+  });
```
htuna07, line 978: Could you add one more test to ensure that we give users enough login chances(3 attempts for this scenario) after they have waited for the block duration?

apps/api/src/main.ts
```diff
@@ -120,6 +120,16 @@ const args = yargs(process.argv.slice(2))
...
+    "passport-identity-failed-login-attempt-limit": {
+      number: true,
+      description: "Maximum failed login attempt before blocking further attempts.",
+      default: 3
+    },
+    "passport-identity-block-duration-after-failed-login-attempts": {
+      number: true,
+      description: "Duration of blocking login attempts in minutes.",
+      default: 30
+    },
```
htuna07: I think we should set these values to zero as default, but I'm not sure that system will work then.

apps/api/src/passport/test/e2e.spec.ts
mavliut-fozliev: This method does not return a response. Should the method body be changed?

htuna07: You can return the response from the login method

apps/api/src/main.ts
mev-isa: Done

apps/api/src/passport/test/e2e.spec.ts
mev-isa: I do not get the "Too many failed login attempts" error when using Promise.all. I think it is because it sends requests in parallel.

apps/api/src/passport/test/e2e.spec.ts
mev-isa: Done

apps/api/src/passport/test/e2e.spec.ts
mev-isa: Done

apps/api/src/passport/test/e2e.spec.ts
mev-isa: Done

apps/api/src/passport/test/e2e.spec.ts
mev-isa: Done

apps/api/src/main.ts
htuna07: Coukd you also set assport-identity-failed-login-attempt-limit to zero as default and test that works fine?

apps/api/src/asset/test/identity.e2e.spec.ts
```diff
@@ -46,6 +46,10 @@ describe("identity-settings", () => {
         issuer: "spica",
         maxExpiresIn: MAX_EXPIRES_IN,
         secretOrKey: "secret",
+        blockingOptions: {
+          blockDurationMinutes: 0,
+          failedAttemptLimit: 100
```
htuna07: It should be zero as default

apps/api/src/function/packages/identity/test/identity.integration.spec.ts
```diff
@@ -38,7 +38,11 @@ describe("Identity", () => {
           defaultIdentityIdentifier: "spica",
           defaultIdentityPassword: "spica",
           audience: "spica",
-          defaultIdentityPolicies: ["PassportFullAccess"]
+          defaultIdentityPolicies: ["PassportFullAccess"],
+          blockingOptions: {
+            blockDurationMinutes: 0,
+            failedAttemptLimit: 100
```
htuna07: should be zero as default

apps/api/src/passport/identity/test/identity.controller.spec.ts
```diff
@@ -33,7 +33,11 @@ describe("Identity Controller", () => {
           expiresIn: 1000,
           issuer: "spica",
           maxExpiresIn: 1000,
-          secretOrKey: "spica"
+          secretOrKey: "spica",
+          blockingOptions: {
+            blockDurationMinutes: 0,
+            failedAttemptLimit: 100
```
htuna07: Should be zero

apps/api/src/passport/test/e2e.spec.ts
```diff
@@ -895,6 +901,92 @@ describe("E2E Tests", () => {
...
+      }
+
+      const responseWithBlockedError = responses[responses.length - 1];
+
+      expect(responseWithBlockedError.statusCode).toEqual(401);
+      expect(responseWithBlockedError.statusText).toEqual("Unauthorized");
+
+      const expectedMessages = [
+        "Too many failed login attempts. Try again after 10 minutes.",
+        "Too many failed login attempts. Try again after 9 minutes 59 seconds."
```
htuna07: They could be failed if tests take more than two seconds, instead of this, you can use jest.setSystemTime to fix the time.

apps/api/src/passport/test/e2e.spec.ts
```diff
@@ -895,6 +901,92 @@ describe("E2E Tests", () => {
...
+      await login();
+      expect(token).toBeUndefined();
+
+      jest.advanceTimersByTime(11 * 60 * 1000);
+
+      await login();
+      expect(token).toBeDefined();
+    });
+
+    it("should lock again", async () => {
```
htuna07: "Should give three chances after the block duration ended" would be better.

apps/api/src/status/test/controller.spec.ts
```diff
@@ -137,6 +137,10 @@ describe("Status", () => {
             maxExpiresIn: 1000,
             secretOrKey: "spica",
             entryLimit: 20,
+            blockingOptions: {
+              failedAttemptLimit: 100,
```
htuna07: Should be zero

htuna07 (changes requested):
Waiting for new changes

apps/api/src/passport/test/e2e.spec.ts
htuna07: Could you try this again, because it shouldn't work like that.

apps/api/src/main.ts
mev-isa: Done

apps/api/src/asset/test/identity.e2e.spec.ts
mev-isa: Done

apps/api/src/function/packages/identity/test/identity.integration.spec.ts
mev-isa: Done

apps/api/src/passport/identity/test/identity.controller.spec.ts
mev-isa: Done

apps/api/src/passport/test/e2e.spec.ts
mev-isa: Done

apps/api/src/passport/test/e2e.spec.ts
mev-isa: Done

apps/api/src/status/test/controller.spec.ts
mev-isa: Done

PR #1330, opened 2025-06-04
test(api): wait until storage indexes are ready

htuna07:


apps/api/src/storage/test/storage.acceptance.spec.ts
```diff
@@ -591,14 +594,13 @@ describe("Storage Acceptance", () => {
...
-          expect(error.response.statusCode).toBe(400);
-          expect(error.response.message).toBe("An object with this name already exists.");
-        });
+      const res = await req.post("/storage", serialize({content: objects}), {
+        "Content-Type": "application/bson"
+      });
+
+      expect(res.statusCode).toEqual(400);
+      expect(res.statusText).toEqual("Bad Request");
+      expect(res.body.message).toEqual("An object with this name already exists.");
```
htuna07, line 603: Tests were silently passing before, since bad request exceptions never caused promise rejection. This way we expect the bad request exception more explicitly and tests will fail if storage accepts inserting two objects with identical names.

PR #1332, opened 2025-06-05
feat(api): prevent re-using old passwords

mavliut-fozliev:


htuna07 (changes requested):
Could you perform the following renamings?
- [ ] passwordHistoryUniquenessCount => passwordHistoryLimit
- [ ] passport-identity-password-history-uniqueness-count => passport-identity-password-history-limit

apps/api/src/passport/identity/src/identity.controller.ts
```diff
@@ -383,13 +408,8 @@ export class IdentityController {
...
     return this.identityService
-      .findOneAndUpdate({_id: id}, {$set: identity}, {returnDocument: ReturnDocument.AFTER})
-      .then(updatedIdentity => {
-        if (!updatedIdentity) {
-          throw new NotFoundException(`Identity with ID ${id} not found`);
-        }
-        return this.afterIdentityUpsert(updatedIdentity);
-      })
+      .findOneAndUpdate({_id: id}, {$set: identity}, {returnDocument: "after"})
+      .then(updatedIdentity => this.afterIdentityUpsert(updatedIdentity))
```
htuna07: Why did we remove the previous code?

apps/api/src/passport/test/e2e.spec.ts
```diff
@@ -976,6 +978,96 @@ describe("E2E Tests", () => {
...
+      expect(errRes.body.message).toEqual("New password can't be the one of last 2 passwords.");
+
+      await updatePassword("brandNewPassword");
+
+      await sleep(1000);
+      await login("spica", "brandNewPassword");
+
+      const reUpdateResponse = await updatePassword("spica");
+      expect(reUpdateResponse.statusCode).toEqual(200);
+    });
```
htuna07: Most of them seem fine, but could you add some expectations after the "await login.." lines to ensure that the login succeeded?

apps/api/src/passport/identity/src/identity.controller.ts
mev-isa: Restored

htuna07 (approved):

PR #1334, opened 2025-06-11
test: fix version control tests

mavliut-fozliev:


apps/api/src/versioncontrol/test/e2e.spec.ts
```diff
@@ -396,12 +411,8 @@ describe("Versioning e2e", () => {
       });
 
       it("should clean", async () => {
-        await insertBucket(getEmptyBucket());
-        await req.post("/versioncontrol/commands/add", {args: ["."]});
-        await req.post("/versioncontrol/commands/rm", {args: ["--cached", "-r", "bucket"]});
-
         const res = await req.post("/versioncontrol/commands/clean", {args: ["-f", "-d"]});
-        expect(res.body.folders).toEqual(["bucket/"]);
+        expect(res.body.folders).toEqual(["bucket/", "function/"]);
```
htuna07, line 415: Why did we remove these lines?

apps/api/src/versioncontrol/test/integration.spec.ts
```diff
@@ -131,43 +131,29 @@ describe("Versioning", () => {
...
-          it("should delete if schema has been deleted", async () => {
-            await prefService.deleteOne({scope: "passport"});
-            await synchronizer.synchronize(SyncDirection.DocToRep);
 
-            // it returns the default schema
-            const file = await rep.readResource("preference", "identity");
-            expect(file).toEqual({
-              _id: "identity",
-              contents: {schema: defaultPref.identity}
-            });
```
htuna07: Why did we remove these tests?

apps/api/src/versioncontrol/test/integration.spec.ts
```diff
@@ -421,6 +434,7 @@ describe("Versioning", () => {
         packages = await engine.getPackages(fn);
         expect(packages).toEqual([]);
 
+        console.log("fn: ", fn);
```
htuna07: Could you remove this line

apps/api/src/versioncontrol/test/representative.spec.ts
```diff
@@ -35,84 +39,118 @@ describe("Representative", () => {
...
-        "module1",
-        "id1",
-        "package",
-        {dependencies: {dep1: "1.1"}},
-        "json"
-      );
-      await representative.write("module1", "id1", "schema", {title: "hi"}, "yaml");
-      await representative.write("module1", "id1", "index", "console.log(123)", "js");
+  describe("watch", () => {
+    let subscribtion: Subscription;
```
htuna07: Fix the typo please

apps/api/src/versioncontrol/test/representative.spec.ts
```diff
@@ -35,84 +39,118 @@ describe("Representative", () => {
...
+
+      setTimeout(() => {
+        const stringified = YAML.stringify({title: "hi"});
+        representative.write("module1", "id1", "schema", stringified, "yaml");
+      }, 1000);
+
+      setTimeout(() => {
+        const updated = YAML.stringify({title: "hello"});
+        representative.write("module1", "id1", "schema", updated, "yaml");
+      }, 2000);
```
htuna07: Do we have to wait almost three seconds in here? Can't we use something like below
await firstPromise
await secondPromise

apps/api/src/versioncontrol/test/representative.spec.ts
```diff
@@ -35,84 +39,118 @@ describe("Representative", () => {
...
-          _id: "id1",
-          contents: {
-            package: {dependencies: {dep1: "1.1"}},
-            schema: {title: "hi"},
-            index: "console.log(123)"
-          }
+    it("should track update change type", done => {
+      subscribtion = representative.watch("module1", ["schema.yaml"]).subscribe({
+        next: change => {
+          if (change.changeType == ChangeTypes.INSERT) return;
```
htuna07: Could you use RxJS operators to skip the first message instead of this if check?

htuna07: Because if the update change comes before the insert, tests will be passed, which should be failed.

libs/representative/src/vcrepresentative.ts
```diff
@@ -54,6 +54,39 @@ export class VCRepresentativeManager implements IRepresentativeManager {
     return fs.promises.rm(dir, {recursive: true});
   }
 
+  // for test use only
```
htuna07: Could you explain why we need this method?

htuna07 (changes requested):
Waiting for changes

apps/api/src/versioncontrol/test/e2e.spec.ts
mev-isa: We used these lines to create directory folders in representative directory, now these folders are created when watchers start watching

apps/api/src/versioncontrol/test/integration.spec.ts
mev-isa: New version control system does not track delete events in the preference module

apps/api/src/versioncontrol/test/integration.spec.ts
mev-isa: Done

apps/api/src/versioncontrol/test/representative.spec.ts
mev-isa: Done

apps/api/src/versioncontrol/test/representative.spec.ts
mev-isa: Done

apps/api/src/versioncontrol/test/representative.spec.ts
mev-isa: I did not use async methods because I call done callback

libs/representative/src/vcrepresentative.ts
mev-isa: This is a convenient method to use in tests

apps/api/src/versioncontrol/test/e2e.spec.ts
htuna07: Does this test ensure that the executed git command is working as expected?

mev-isa: Previously I added a new bucket and removed the changes so that the bucket folder remained, now it is created immediately at the beginning and the clean command deletes it as before

apps/api/src/versioncontrol/test/integration.spec.ts
htuna07: But why? What will happen if it's deleted?

htuna07: Also, insert tests have been removed, shouldn't we track it too?

apps/api/src/versioncontrol/test/representative.spec.ts
htuna07: Okay then can't you use the following approach?
firstPromise.then(()=>secondPromise)

libs/representative/src/vcrepresentative.ts
htuna07: I can see that clearly, I'm just wondering why you needed to write this kind of method here for only tests? Can't you do the same with existing methods?

apps/api/src/versioncontrol/test/representative.spec.ts
mev-isa: Done

libs/representative/src/vcrepresentative.ts
mev-isa: Done

htuna07 (approved):

PR #1338, opened 2025-06-12
feat(cli): add ability to filter resources by ID

mavliut-fozliev:


apps/cli/src/commands/project/sync.ts
```diff
@@ -65,10 +77,6 @@ async function sync({
   for (const name of modules) {
     const moduleSynchronizers = synchronizers.filter(s => s.moduleName == name);
 
-    if (!moduleSynchronizers.length) {
-      return Promise.reject(`Module ${name} does not exist.`);
-    }
-
```
htuna07, line 71: Why did we remove this check?

apps/cli/src/commands/project/sync.ts
```diff
@@ -241,13 +273,18 @@ export class FunctionSynchronizer implements ModuleSynchronizer {
...
     private targetService: httpService.Client,
-    private options: {syncFnEnv}
+    private options: {syncFnEnv; functionIds}
   ) {}
 
+  async getFunctions(service: httpService.Client) {
+    const allFunctions = await service.get<any[]>("function");
+    return getFilteredResources(allFunctions, this.options.functionIds);
+  }
+
```
htuna07: Is it possible to filter resources on the API side instead of manually filtering?

apps/cli/src/validator.ts
```diff
@@ -40,6 +40,13 @@ ${availableSyncModules.map(m => `- ${green(m)}`).join("\n")}
   return input;
 }
 
+export function validateSyncIds(input: any): string[] {
+  if (typeof input !== "string") {
+    throw new Error("You should provide a comma separated string");
+  }
+  return input.split(",");
+}
+
```
htuna07: There is a manipulation on the last line that violates the single responsibility rule for this function.

apps/cli/test/commands/project/sync.spec.ts
```diff
@@ -139,7 +139,9 @@ describe("Synchronize", () => {
       });
 
       describe("BucketSynchronizer", () => {
-        const synchronizer = new BucketSynchronizer(sourceService as any, targetService as any);
+        const synchronizer = new BucketSynchronizer(sourceService as any, targetService as any, {
+          bucketIds: undefined
+        });
```
htuna07, line 144: Could you also add new tests for filtering resources?

apps/cli/src/commands/project/sync.ts
mev-isa: Since it throws the wrong error, like when we provide non-existent bucket ID, it fails to initialize bucket data synchronizer and throws this error. Also, we have a check method that checks the provided module names. 
I can return if this is expected behavior or change the error text/

apps/cli/src/commands/project/sync.ts
mev-isa: I couldn't filter buckets, so I did it all in one style.

apps/cli/src/validator.ts
mev-isa: Done

apps/cli/test/commands/project/sync.spec.ts
mev-isa: Done

htuna07 (approved):

PR #1342, opened 2025-06-17
deprecate(cli): deprecate --sync-fn-env

Mozan0:


htuna07 (changes requested):
We should also remove the implementation of the sync-fn-env flag.

Mozan0:
> We should also remove the implementation of the sync-fn-env flag.

Done

PR #1343, opened 2025-06-17
feat(api): handle replica apis for new version control

mev-isa:


libs/interface/versioncontrol/src/synchronizer.ts
```diff
@@ -106,21 +107,40 @@ export type SynchronizerArgs<R1 extends Resource, R2 extends Resource> = {
...
+  commander?: ICommander;
 };
 
 export abstract class Synchronizer<R1 extends Resource, R2 extends Resource> {
-  constructor(private args: SynchronizerArgs<R1, R2>) {}
+  constructor(private args: SynchronizerArgs<R1, R2>) {
+    if (args.commander) {
+      args.commander.register(
+        this,
+        [this.addDocToRepAction, this.addRepToDocAction],
```
htuna07: Shouldn't we also need to define and register **remove** methods?

htuna07 (commented):
Waiting for final change.

libs/interface/versioncontrol/src/synchronizer.ts
mev-isa: Done

PR #1347, opened 2025-06-23
test(api): separate bucket-data controller tests

mavliut-fozliev:


htuna07 (commented):
Could you also move other bucket-data test files to this folder? Moved files shouldn't start with bucket-data by the way.

PR #1353, opened 2025-06-26
fix(firehose): seperate socket management from enqueue

Mozan0:


apps/api/src/function/enqueuer/src/firehose.ts
```diff
@@ -103,33 +103,40 @@ export class FirehoseEnqueuer extends Enqueuer<FirehoseOptions> {
...
+    }
+
     for (const pair of this.eventTargetPairs) {
       if (
-        pair.name == name ||
-        pair.name == "*" ||
-        (pair.name == "**" && (name == "connection" || name == "close"))
+        pair.name === name ||
+        pair.name === "*" ||
+        (pair.name === "**" && (name === "connection" || name === "close"))
```
htuna07: If it's not necessary, could you revert strict equality checks(===) back?

Mozan0: Done

htuna07 (approved):

htuna07 (approved):

PR #1355, opened 2025-06-30
 feat(ui): update login page to match pixel-perfect Figma design

limonatasuyu:


.yarn/install-state.gz
Ayazelar, line 1: Remove this file from changes. it's not connected to this task

.yarnrc.yml
Ayazelar, line 1: Remove this file from changes. it's not connected to this task

apps/panel/src/pages/login/Login.module.scss
```diff
@@ -9,31 +9,79 @@
...
+      outline: none;
+    }
+
+    &:focus-within, &:focus {
+      border-color: rgb(var(--color-primary-rgb));
+    }
+
+    input,
+    span {
+      line-height: var(--font-size-sm) !important;
```
Ayazelar: Use the line-height variable. If it does not exist in root ,create a new one

apps/panel/src/pages/login/Login.module.scss
```diff
@@ -9,31 +9,79 @@
...
+    }
+
+  }
+
+  .submitButton {
+    background-color: var(--color-primary) !important;
+  }
+
+  .formButtonContainer {
+    height: 16px !important;
```
Ayazelar: no need to give static 16px. cause there is nothing to have same height to 16

yarn.lock
Ayazelar, line 1: Remove this file from changes. it's not connected to this task

apps/panel/src/pages/login/Login.module.scss
```diff
@@ -9,31 +9,79 @@
...
+    font-size: var(--font-size-sm);
+    font-family: var(--font-family-base);
+  }
+
+  .formButton {
+    padding-bottom: calc(var(--padding-sm) + var(--padding-xs));
+    padding-top: calc(var(--padding-sm) + var(--padding-xs));
+  }
```
Ayazelar: no need to change padding. It should come from the size prop in the button

PR #1357, opened 2025-07-01
feat: add reusable useLocalStorage hook with full TypeScript support

limonatasuyu:
- Includes a sample component demonstrating hook usage


apps/panel/src/custom-hooks/useLocalStorage.ts
Ayazelar, line 1: change folder name (hooks/useLocalStorage.ts)

Ayazelar (changes requested):
Examples are intended for local testing only and should not be included in production code or the pull request.

PR #1358, opened 2025-07-01
feat: create AuthContext and useAuth Hook for Authentication Management

limonatasuyu:


.yarn/install-state.gz
Ayazelar, line 1: remove this change. its not connected to this task

.yarnrc.yml
Ayazelar, line 1: remove this change. its not connected to this task

apps/panel/src/contexts/AuthContext.tsx
```diff
@@ -0,0 +1,101 @@
...
+            setUserToken(null);
+        } finally {
+            setLoading(false);
+        }
+    }, [userToken, setUserToken]);
+
+    useEffect(() => {
+        if (user) return;
+        fetchUser();
+    }, [user, fetchUser]);
```
Ayazelar: Token validation should be handled outside of the fetchUser function. This makes fetchUser cleaner and more focused, improving overall readability.
For example, the useEffect can check for the token before calling fetchUser, instead of having that logic inside the function.

apps/panel/src/contexts/AuthContext.tsx
```diff
@@ -0,0 +1,101 @@
...
+            console.error('Login error:', error);
+        } finally {
+            setLoading(false);
+        }
+    }, [setUserToken]);
+
+    const logout = useCallback(() => {
+        setUserToken(null);
+        setUser(null);
+    }, [setUserToken]);
```
Ayazelar: In the current implementation, setUserToken and setUser are included in the dependency array of the logout function, even though they are stable functions returned from useState.

React guarantees that setState functions never change between renders, so there's no need to include them in the dependency list.

Ayazelar: also apply for others

apps/panel/src/types/user.ts
Ayazelar, line 1: isn't that type? why are you used interface

apps/panel/src/custom-hooks/useLocalStorage.tsx
Ayazelar, line 1: When useLocalStorage pr is merged, this should be updated. it should be included among the changes that will not be recreated and added.

PR #1366, opened 2025-07-07
feat: Create a reusable useApi hook for future API requests

limonatasuyu:


Ayazelar (changes requested):
Instead of triggering the API call on mount, we should refactor the useApi hook to expose a request method. This way, the consumer can control when the API call happens, making it more flexible and reusable across different scenarios (e.g., conditional calls, retries, etc.).

Also, please ensure:
1. The base URL is appended only if the given URL is relative.
2. The Authorization header is automatically added if a token exists in localStorage.
3. defaultBaseUrl should come from env file

apps/panel/src/hooks/useApi.ts
```diff
@@ -0,0 +1,67 @@
+import axios from "axios";
+import {useCallback, useEffect, useMemo, useState} from "react";
+import useLocalStorage from "./useLocalStorage";
+
+type ApiRequestOptions = {
+  endpoint: string;
+  method?: "get" | "post" | "put" | "delete";
```
Ayazelar: add patch

apps/panel/src/hooks/useApi.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const request = useCallback(
+    (body?: any) => {
+      const makeRequest = async () => {
+        try {
+          const response = await axios({
+            method,
+            url: resolvedUrl,
+            data: body,
+            headers: token ? {Authorization: token} : undefined
```
Ayazelar: Allow users to pass custom headers to the API request function. These headers should be added first to the headers object, so they can override defaults if needed. Also, if a token is provided and no custom Authorization header exists, add Authorization: IDENTITY ${token} automatically.

Ayazelar (approved):

PR #1371, opened 2025-07-08
feat(migrate): add new bucket index options

Mozan0:


apps/migrate/src/migrations/0.16.0/update_old_index_usage_to_new.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (indexes.length > 0) {
+      updateOp.updateOne.update.$set = {
+        indexes,
+        properties: bucket.properties
+      };
+    } else {
+      updateOp.updateOne.update.$set = {
+        properties: bucket.properties
+      };
+    }
```
htuna07: It seems the code below also runs and is shorter
```suggestion
    const updateOp: any = {
      updateOne: {
        filter: {_id: bucket._id},
        update: {
            $set: {
                properties: bucket.properties,
                indexes
            }
        }
      }
    };
```

Mozan0: Done.

PR #1372, opened 2025-07-08
feat: create authService with useApi, integrate into login page

limonatasuyu:


apps/panel/src/services/authService.ts
```diff
@@ -0,0 +1,55 @@
...
+      requestLogin({body: {identifier, password}});
+    },
+    [requestLogin]
+  );
+
+  useEffect(() => {
+    if (!loginData) return;
+    setToken(loginData.token);
+    navigate("/home");
+  }, [loginData, navigate, setToken]);
```
Ayazelar: remove useEffect.

apps/panel/src/pages/login/Login.tsx
```diff
@@ -58,9 +45,18 @@ const Login = () => {
...
           >
             <Logo size="xl" />
-
-            {loginError && <div className={styles.errorText}>{loginError}</div>}
-
+            {loginLoading && (
+              <Text className={styles.loadingText} size="medium">
+                Logging in...
+              </Text>
+            )}
```
Ayazelar: no need to show text. button is disabled when the loading is true

Ayazelar (changes requested):
.

limonatasuyu:
changes made

PR #1374, opened 2025-07-08
feat: implement BucketContext and bucketService

limonatasuyu:


apps/panel/src/contexts/BucketContext.tsx
```diff
@@ -0,0 +1,37 @@
+import {createContext, useMemo, useContext, type ReactNode, useEffect} from "react";
+import {useBucketService} from "../services/bucketService";
+
+type BucketContextType = {
+  buckets: any; // Replace with actual type
```
Ayazelar: Do not use any. give type

Ayazelar (changes requested):
.

apps/panel/src/contexts/BucketContext.tsx
limonatasuyu: done.

Ayazelar (approved):

PR #1378, opened 2025-07-10
refactor(api): move duplicated handle realtime functions to libs

Mozan0:


libs/realtime/src/gateway.ts
```diff
@@ -0,0 +1,78 @@
...
+      catchError(error => {
+        const errMsg = buildErrorMessage(error);
+        client.send(JSON.stringify(errMsg));
+        client.close(1003);
+        return of(null);
+      })
+    );
+
+    stream.pipe(takeUntil(fromEvent(client, "close"))).subscribe(data => {
+      if (data !== null) {
```
htuna07, line 58: Is this check necessary?

Mozan0: Yes, since stream catchError returns null and second part emits data from there this if control becomes reasonable.

`const stream = realtime.find(collection, options).pipe(
      catchError(error => {
        const errMsg = buildErrorMessage(error);
        client.send(JSON.stringify(errMsg));
        client.close(1003);
        return of(null); // This returns null
      })
    )`

PR #1387, opened 2025-07-14
feat: read token from localStorage and decode identifier for name prop and pass to toolbar in Layout.tsx

limonatasuyu:


apps/panel/src/components/atoms/toolbar/Toolbar.tsx
```diff
@@ -3,15 +3,16 @@ import React, {memo, useState, type FC} from "react";
 import styles from "./Toolbar.module.scss";
 
 type TypeToolbar = {
-  token: string;
-  name: string;
+  token: string | null;
+  name: string | null;
```
Ayazelar:
```suggestion
  token: string;
  name: string;
```

Ayazelar: There is no need to add the null type

apps/panel/src/components/atoms/toolbar/Toolbar.tsx
```diff
@@ -3,15 +3,16 @@ import React, {memo, useState, type FC} from "react";
...
   onDrawerOpen?: () => void;
 };
 
 const Toolbar: FC<TypeToolbar> = ({token, name, onDrawerOpen}) => {
   const [copied, setCopied] = useState(false);
 
   const handleCopy = () => {
+    if (!token) return;
```
Ayazelar:
```suggestion
```

apps/panel/src/layout/Layout.tsx
```diff
@@ -24,6 +30,14 @@ const Layout = () => {
     return () => window.removeEventListener("resize", handleResize);
   }, [isDrawerOpen]);
 
+
+  useEffect(() => {
+    if (!token || !token.length) return;
+    const decoded = jwtDecode<AuthTokenJWTPayload>(token);
+    const name = decoded.identifier
+    setName(name)
+  }, [token])
```
Ayazelar: no need to use useEffect use variable directly

Ayazelar (changes requested):
.

PR #1403, opened 2025-07-22
feat(auth): handle 401s with observer, add localStorage sync and delete support

limonatasuyu:
- useRequestTracker redirects on 401 responses
- useLocalStorage updates value on same-tab changes
- deleteValue utility added to clear stored key
- AuthContext and Layout updated to use new logic


Ayazelar (changes requested):
There should only be task-related changes.

PR #1404, opened 2025-07-23
feat(api): introduce resumable uploads for storage

mavliut-fozliev:
EXAMPLE CLIENTS

Nodejs:

```
import fs from "fs";
import tus from "tus-js-client";

const filename = "Docker.dmg";
const filePath = "./" + filename;
const fileSize = fs.statSync(filePath).size;
const fileStream = fs.createReadStream(filePath);

const upload = new tus.Upload(fileStream, {
  endpoint: "http://localhost:4300/storage/resumable",
  metadata: {
    filename,
    filetype: "application/x-apple-diskimage",
  },
  uploadSize: fileSize,
  urlStorage: new tus.FileUrlStorage("./tus-urls.json"), // <= save URL
  onError(error) {
    console.error("❌ Upload failed:", error);
  },
  onProgress(bytesUploaded, bytesTotal) {
    const percentage = ((bytesUploaded / bytesTotal) * 100).toFixed(2);
    console.log(`📤 Uploaded ${percentage}%`);
  },
  onSuccess() {
    console.log("✅ Upload finished:", upload.url);
  },
});

upload.findPreviousUploads().then((previousUploads) => {
  if (previousUploads.length) {
    upload.resumeFromPreviousUpload(previousUploads[0]);
  }
  upload.start();
});

```

Browser:

```
import tus from "tus-js-client";

const input = document.querySelector('input[type="file"]');

input.addEventListener('change', () => {
  const file = input.files[0];
  if (!file) return;

  const upload = new tus.Upload(file, {
    endpoint: "http://localhost:4300/storage/resumable",
    metadata: {
      filename: file.name,
      filetype: file.type || "application/octet-stream",
    },
    // In the browser, tus-js-client determines the file size itself, uploadSize is not needed
    // In the browser, urlStorage works automatically (uses localStorage)
    onError(error) {
      console.error("❌ Upload failed:", error);
    },
    onProgress(bytesUploaded, bytesTotal) {
      const percentage = ((bytesUploaded / bytesTotal) * 100).toFixed(2);
      console.log(`📤 Uploaded ${percentage}%`);
    },
    onSuccess() {
      console.log("✅ Upload finished:", upload.url);
    },
  });

  upload.findPreviousUploads().then((previousUploads) => {
    if (previousUploads.length) {
      upload.resumeFromPreviousUpload(previousUploads[0]);
    }
    upload.start();
  });
});

```

### DIFF
--- a/apps/api/src/main.ts
+++ b/apps/api/src/main.ts
@@ -287,6 +287,11 @@ const args = yargs(process.argv.slice(2))
     "storage-total-size-limit": {
       number: true,
       description: "Total size limit of storage. Unit: Mb"
+    },
+    "expiration-period": {
+      number: true,
+      description: "Storage period for unloaded files in milliseconds, default is 2 days",
+      default: 1000 * 60 * 60 * 24 * 2 // 2 days
     }
   })
   /* Status Options */
@@ -516,7 +521,8 @@ const modules = [
     awss3CredentialsPath: args["awss3-credentials-path"],
     awss3BucketName: args["awss3-bucket-name"],
     objectSizeLimit: args["storage-object-size-limit"],
-    totalSizeLimit: args["storage-total-size-limit"]
+    totalSizeLimit: args["storage-total-size-limit"],
+    expirationPeriod: args["expiration-period"]
   }),
   PassportModule.forRoot({
     publicUrl: args["public-url"],
--- a/apps/api/src/main.ts
+++ b/apps/api/src/main.ts
@@ -288,7 +288,7 @@ const args = yargs(process.argv.slice(2))
       number: true,
       description: "Total size limit of storage. Unit: Mb"
     },
-    "expiration-period": {
+    "resumable-upload-expires-in": {
       number: true,
       description: "Storage period for unloaded files in milliseconds, default is 2 days",
       default: 1000 * 60 * 60 * 24 * 2 // 2 days
@@ -522,7 +522,7 @@ const modules = [
     awss3BucketName: args["awss3-bucket-name"],
     objectSizeLimit: args["storage-object-size-limit"],
     totalSizeLimit: args["storage-total-size-limit"],
-    expirationPeriod: args["expiration-period"]
+    resumableUploadExpiresIn: args["resumable-upload-expires-in"]
   }),
   PassportModule.forRoot({
     publicUrl: args["public-url"],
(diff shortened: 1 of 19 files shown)


mev-isa:
Current state:

local storage:
✅ working
✅ expirations are deleted

s3:
✅ working
✅ expirations are deleted

gcloud:
✅ working
✅ expirations are deleted (custom implementation)

apps/api/src/main.ts
```diff
@@ -287,6 +287,11 @@ const args = yargs(process.argv.slice(2))
     "storage-total-size-limit": {
       number: true,
       description: "Total size limit of storage. Unit: Mb"
+    },
+    "expiration-period": {
+      number: true,
+      description: "Storage period for unloaded files in milliseconds, default is 2 days",
+      default: 1000 * 60 * 60 * 24 * 2 // 2 days
```
htuna07: Could you use something like "resumable-upload-expires-in"? The description could be cleaner.
You should also update the "expirationPeriod" names in the system.

apps/api/src/storage/src/strategy/aws.s3.ts
```diff
@@ -88,4 +94,31 @@ export class AWSS3 implements Strategy {
...
+          accessKeyId: config.accessKeyId,
+          secretAccessKey: config.secretAccessKey
+        }
+      },
+      expirationPeriodInMilliseconds: this.expirationPeriod
+    });
+  }
+
+  async getFileInfo(id: string) {
+    const infoBuffer = await this.read(id + ".info");
```
htuna07: If Tus has a method to read info, it should have been preferred.

apps/api/src/storage/src/storage.service.ts
```diff
@@ -245,4 +274,32 @@ export class StorageService extends BaseCollection<StorageObjectMeta>("storage")
...
+      }
+    };
+
+    try {
+      await this._coll.insertOne(document);
+    } catch (exception) {
+      throw new BadRequestException(
+        exception.code === 11000 ? "An object with this name already exists." : exception.message
+      );
+    }
```
htuna07: We should remove data from disk to keep the database synchronized.

apps/api/src/main.ts
mev-isa: Done

apps/api/src/storage/src/strategy/aws.s3.ts
mev-isa: Done

apps/api/src/storage/src/storage.service.ts
mev-isa: Done

htuna07 (approved):

PR #1409, opened 2025-07-28
fix(api): handle directory create&delete for default storage strategy

Mozan0:


htuna07 (changes requested):
Seems okay, but it's better to move some copy-pasted code to methods like
- Define a isDirectory method for "id.endswith" check, return boolean
- Define a createDir and removeDir methods for mkdir and remove calls of fs.promises
Structure could be the same(if, return lines).

PR #1427, opened 2025-08-06
feat(api): add simpleActionGuard for flexible resource filtering

Mozan0:


apps/api/src/passport/guard/src/action.guard.ts
```diff
@@ -349,3 +349,67 @@ function createActionGuard(
...
+          if (statement.resource.include) includedResources.push(...statement.resource.include);
+          if (statement.resource.exclude) excludedResources.push(...statement.resource.exclude);
+        }
+      }
+
+      request.resourceFilter = {include: includedResources, exclude: excludedResources};
+
+      if (matchedStatements.length > 0) {
+        return true;
+      }
```
htuna07: Could you move this setting resource filter into the if line, otherwise it makes no sense to set the resource filter even if we throw a forbidden exception.
We should also expect the length of includedResources to be greater than zero.
Early return would be better for the error case, by the way.

apps/api/src/passport/guard/test/simple.action.guard.spec.ts
```diff
@@ -0,0 +1,472 @@
...
+      });
+
+      expect(await guard()).toBe(true);
+      expect(request.resourceFilter).toEqual({
+        include: [],
+        exclude: ["bucket1/row2"]
+      });
+    });
+  });
+});
```
htuna07, line 571: Could you also add edge case test for scenarios like the include is not an array? like string/number etc?

apps/api/src/passport/guard/test/simple.action.guard.spec.ts
```diff
@@ -0,0 +1,472 @@
...
+          }
+        ]
+      });
+
+      expect(await guard()).toBe(true);
+      expect(request.resourceFilter).toEqual({
+        include: ["*/*"],
+        exclude: ["bucket2/*"]
+      });
+    });
```
htuna07, line 204: This test is fine, but I'm also wondering what if the **include** array has a single wildcard(*) while the resource i try to access has two params like bucket and row? We should expect the resource filter include array should be single wildcard

htuna07 (approved):
Seems okay

apps/api/src/passport/guard/src/action.guard.ts
Mozan0: done

apps/api/src/passport/guard/test/simple.action.guard.spec.ts
Mozan0: done

apps/api/src/passport/guard/test/simple.action.guard.spec.ts
Mozan0: done

htuna07 (dismissed):

htuna07 (dismissed):

PR #1429, opened 2025-08-07
feat(api): allow get by object name in storage

Mozan0:


apps/api/src/storage/src/storage.controller.ts
```diff
@@ -74,10 +75,15 @@ export class StorageController {
   @UseGuards(AuthGuard(), ActionGuard("storage:show", "storage/:id"))
   async view(
     @Res() res,
-    @Param("id", OBJECT_ID) id: ObjectId,
+    @Param("id", OR(v => /^[a-f\d]{24}$/i.test(v), OBJECT_ID)) idOrName: ObjectId | string,
```
htuna07: Why don't you use some method like isObjectId from @spica-server/database?

PR #1447, opened 2025-08-22
test(api): add function controller tests

Mozan0:


apps/api/src/function/test/controller.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+    const body = await request
+      .get("/function", {filter: JSON.stringify({index: "findMe("})})
+      .then(r => r.body);
+    expect(body.statusCode).toEqual(400);
+    expect(body.error).toEqual("Bad Request");
+    expect(body.message).toContain("Invalid regular expression");
+  });
+
+  describe("endpoints", () => {
+    it("should return information about enqueuers/runtimes and timeout", async () => {
```
htuna07, line 118: Is it possible to expect more specific results in this test instead of just arrays?

apps/api/src/function/test/controller.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+      const found = await request.get(`/function/${inserted._id}`).then(r => r.body);
+      expect(found).toEqual({...inserted, env_vars: []});
+    });
+
+    it("should delete a function and return 204", async () => {
+      const inserted = await request.post("/function", fnSchema).then(r => r.body);
+      const del = await request.delete(`/function/${inserted._id}`);
+      expect(del.statusCode).toEqual(204);
+
+      const getRes = await request.get(`/function/${inserted._id}`);
```
htuna07, line 143: Should it return 404 instead?

apps/api/src/function/test/controller.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+      expect([postRes.statusCode, postRes.statusText]).toEqual([204, "No Content"]);
+
+      const show = await request.get(`/function/${inserted._id}/index`).then(r => r.body);
+      expect(show.index).toContain("hello");
+    });
+
+    it("should return dependencies array (empty by default)", async () => {
+      const inserted = await request.post("/function", fnSchema).then(r => r.body);
+      const deps = await request.get(`/function/${inserted._id}/dependencies`).then(r => r.body);
+      expect(Array.isArray(deps)).toBe(true);
```
htuna07, line 196: If so, why don't we expect empty array?

Mozan0: done

Mozan0:
> Seems nice but there are some missing parts. For example dependency install, uninstall, environment variable resolve, unresolve?

done

apps/api/src/function/test/controller.spec.ts
Mozan0: Need to talk about it

apps/api/src/function/test/controller.spec.ts
Mozan0: done

PR #1453, opened 2025-08-26
feat(api): introduce mailer module

Mozan0:


apps/api/src/main.ts
```diff
@@ -542,6 +571,15 @@ const modules = [
   EnvVarModule.forRoot({
     realtime: args["env-var-realtime"]
   }),
+  MailerModule.forRoot({
+    host: args["mailer-host"],
+    port: args["mailer-port"],
+    secure: args["mailer-secure"],
+    auth: {
+      user: args["mailer-user"],
+      pass: args["mailer-pass"]
```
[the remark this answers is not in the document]
Mozan0: done

apps/api/src/mailer/src/mailer.service.ts
```diff
@@ -0,0 +1,36 @@
...
+      subject: mail.subject,
+      text: mail.text,
+      html: mail.html
+    };
+
+    try {
+      const result = await this.transporter.sendMail(message);
+      return result;
+    } catch (error) {
+      this.logger.error(`Failed to send mail to ${mail.to}: ${error.message}`);
```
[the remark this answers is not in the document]
Mozan0: done

.github/workflows/ci.yml
```diff
@@ -39,6 +39,8 @@ jobs:
           - command: test-image:api --build-arg PROJECTS=api/dashboard/**
```
htuna07, line 39: Mailhog is just a service we use for emails. We should prefer a name like "mail" or "smtp" instead, because we might want to use a different service for emails. Otherwise, we would change all MailHog names to "mail-server-x", for example.

libs/interface/mailer/src/options.ts
```diff
@@ -0,0 +1,14 @@
+export interface MailerOptions {
```
htuna07, line 1: Could you confirm that optional fields are really optional?

libs/interface/mailer/src/service.ts
```diff
@@ -0,0 +1,7 @@
+export interface SendMailOptions {
```
htuna07, line 1: Could you confirm that optional fields are really optional?

apps/api/src/mailer/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "../../tsconfig.json",
+  "include": [
+    "src/**/*.ts",
+    "index.ts",
+    "src/**/*.json",
+    "../../../../libs/interface/mailer/index.ts"
+  ],
```
htuna07: I don't think that we need all of these files. Check other examples to fix here.

apps/api/src/mailer/project.json
```diff
@@ -0,0 +1,5 @@
+{
```
htuna07, line 1: We need to create the same one for tests(names will end with /test), check other examples for more.

apps/api/src/mailer/test/mailer.integration.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+    const extracted = {
+      to: toHeader ? String(toHeader[0]) : String(raw || ""),
+      subject: subjectHeader ? String(subjectHeader[0]) : String(raw || ""),
+      text: String(item.Content?.Body || raw || "")
+    };
+
+    const expected = expect.objectContaining({
+      to: expect.stringContaining("recipient@example.com"),
+      subject: expect.stringContaining("Integration Test"),
+      text: expect.stringContaining("Hello from integration test")
```
htuna07: If not necessary, could you consider removing stringContaining parts? It seems we should expect exactly what we sent.

apps/api/src/mailer/test/mailer.integration.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+        console.error(e);
+      }
+    }
+
+    const options = {
+      host: smtpHost,
+      port: smtpPort,
+      secure: false,
+      defaults: {
+        from: "integration@example.com"
```
htuna07, line 46: Could you test the **from** part too?

.github/workflows/ci.yml
Mozan0: done

apps/api/src/mailer/project.json
Mozan0: done

apps/api/src/mailer/tsconfig.json
Mozan0: done

libs/interface/mailer/src/service.ts
Mozan0: done

libs/interface/mailer/src/options.ts
Mozan0: done

apps/api/src/mailer/test/mailer.integration.spec.ts
Mozan0: done

apps/api/src/mailer/test/mailer.integration.spec.ts
Mozan0: done

Mozan0:
[[SPICA - 722] : Introduce mail service](https://trello.com/c/oog1cl94/722-spica-722-introduce-mail-service)

htuna07 (approved):

PR #1463, opened 2025-09-12
feat(bucket-more-popup): implement Limitations Checkbox and Configure Limitations Drawer

limonatasuyu:


apps/panel/src/contexts/BucketContext.tsx
```diff
@@ -42,19 +42,19 @@ type BucketContextType = {
...
     newField: Property,
     requiredField?: string,
     primaryField?: string
   ) => Promise<any>;
+  updateBucketLimitation: (bucket: BucketType) => Promise<void>;
+  updateBucketLimitationFields: (
+    bucket: BucketType,
+    countLimit: number,
+    limitExceedBehaviour: "prevent" | "remove"
+  ) => Promise<any>;
```
kenangsmv: Type the return value, instead of any

conan-jsx (approved):

PR #1466, opened 2025-09-15
Fix: Ensure Table Width Always Expands to Full Screen

limonatasuyu:


apps/panel/src/components/organisms/table/Table.tsx
```diff
@@ -125,7 +126,27 @@ function getCalculatedColumnWidth(columns: TypeDataColumn[], containerWidth: num
 }
 
 const getFormattedColumns = (containerWidth: number, columns: TypeDataColumn[]) => {
-  const defaultColumnWidth = getCalculatedColumnWidth(columns, containerWidth + 10);
+  const allColumnsHaveWidth =
+    columns.filter(col => col.width !== undefined).length === columns.length;
+
```
kenangsmv, line 131: Check if column length is null.

apps/panel/src/components/organisms/table/Table.tsx
```diff
@@ -125,7 +126,27 @@ function getCalculatedColumnWidth(columns: TypeDataColumn[], containerWidth: num
...
+    const resizableColumns = columns.filter(col => !col.fixedWidth);
+
+    const fixedWidthTotal = fixedWidthColumns.reduce(
+      (total, col) => total + parseWidth(col.width as string, containerWidth),
+      0
+    );
+
+    const remainingWidth = containerWidth - fixedWidthTotal;
+    const widthPerResizableColumn = remainingWidth / resizableColumns.length;
+
```
kenangsmv, line 145: resizableColumns.length could be zero. Dividing by zero will be a problem.
Also, fixedWithTotal could be greater than containerWidth. So "remainingWidth" could be negative number. Maybe Clamp with 0

conan-jsx (changes requested):

conan-jsx (approved):

PR #24, opened 2019-07-16
feat(storage): support for multi upload on storage

tthiscafer:

PR #1847, opened 2026-06-12
fix(schema): stop leaking change-stream subscriptions in schema resolution

htuna07:
## Problem

`Validator._fetch` (`packages/core/schema/src/validator.ts`) is AJV's `loadSchema` callback. When a URI resolver returns an observable (as `BucketSchemaResolver` does), it:

```ts
result.pipe(skip(1), tap(...)).subscribe();   // keep-fresh: never stored, never unsubscribed
return result.pipe(take(1)).toPromise();       // second, independent subscription
```

`BucketSchemaResolver`'s `watchBucket` / `watchPreferences` are **cold** observables — each `subscribe()` opens a fresh MongoDB change-stream cursor. So every `_fetch`:

- opens **two** cursors (keep-fresh + `take(1)`), and
- **permanently leaks** the keep-fresh one (never unsubscribed).

Because the keep-fresh `tap` calls `removeSchema(uri)` on every schema change, the URI gets evicted and the next validation re-runs `_fetch`, leaking yet another cursor. Cursors accumulate over uptime — a change-stream / memory leak.

## Fix

Port of fintech fix #24:

- Wrap the resolved observable in `shareReplay({bufferSize: 1, refCount: true})` so the keep-fresh and `take(1)` consumers **share one** upstream cursor.
- Track the keep-fresh subscription per URI in `_schemaSubscriptions`; unsubscribe the prior one before re-subscribing.
- Tear it down in `removeSchema`.
- Share `preferenceWatcher` via `shareReplay(1)` so its cursor is reused across all bucket resolutions instead of one-per-bucket.

## Evidence (production mongod slow-query logs)

The affected deployment held **~76 concurrent `spica.buckets` change-stream cursors over just 23 watched buckets**, plus **43 `spica.preferences` cursors**, versus **~63 / 32** on the build that already carried this fix. That's **~0.25 cpu-cores/sec** of avoidable change-stream polling — and it grows with uptime since the leak is unbounded.

This is complementary to #1845 (which removed the per-request `buckets.findOne` schema-read storm); that fixed reads, this fixes the watcher-cursor leak.

## Tests

- New regression test in `validator.spec.ts` asserts repeated re-fetch of the same URI opens **4** cursors (one per `_fetch`) and leaves **0** active after `removeSchema`. Against the unpatched validator it opens **8** and leaks **4** — i.e. the test fails without the fix.
- `core/schema`: 28/28 pass.
- `api/bucket` `bucket.schema.resolver.spec.ts`: passes (change propagation through `shareReplay` intact).

### DIFF
--- a/packages/api/bucket/src/bucket.schema.resolver.ts
+++ b/packages/api/bucket/src/bucket.schema.resolver.ts
@@ -3,7 +3,7 @@ import {BucketService, compile} from "@spica-server/bucket-services";
 import {CodeKeywordDefinition, KeywordCxt, Validator, _} from "@spica-server/core-schema";
 import {ObjectId} from "@spica-server/database";
 import {combineLatest, Observable, Subject} from "rxjs";
-import {map, takeUntil} from "rxjs/operators";
+import {map, shareReplay, takeUntil} from "rxjs/operators";
 import {Bucket, BucketPreferences} from "@spica-server/interface-bucket";
 
 @Injectable()
@@ -13,7 +13,7 @@ export class BucketSchemaResolver implements OnModuleDestroy {
   onDestroySubject = new Subject();
 
   constructor(private bucketService: BucketService) {
-    this.preferenceWatcher = this.bucketService.watchPreferences(true);
+    this.preferenceWatcher = this.bucketService.watchPreferences(true).pipe(shareReplay(1));
   }
 
   onModuleDestroy() {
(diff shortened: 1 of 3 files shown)


htuna07:
@claude fix the "Bug: removeSchema() without argument orphans all subscriptions" you found.

idriska (approved):

PR #1855, opened 2026-06-17
fix(passport): enforce 2FA from persisted record to fix multi-replica bypass

htuna07:
## Problem

TOTP/2FA was inconsistent across replicas. When the API runs with two or more replicas, activating 2FA on one replica and then logging in against **another** replica would let the user log straight in **without** a challenge — a 2FA bypass.

### Root cause

`PassportIdentityController.signIdentity` / `PassportUserController.signUser` decided whether to enforce 2FA from the in-memory `AuthFactor.userFactor` map via `hasFactor(id)`:

```ts
if (this.authFactor.hasFactor(id)) { /* challenge */ } else { /* log in */ }
```

That map is **per-replica**. It is populated by a one-time startup load and by live `register()`/`start()` calls propagated to other replicas only via best-effort `ClassCommander` SYNC broadcasts over a MongoDB change-stream bus — which is **live-only (no replay), has a 5-minute TTL, and is a no-op when replication is disabled or the change stream wasn't active**.

So a replica that was already running when 2FA got activated elsewhere — and missed the `register` broadcast — has `hasFactor(id) === false` and skips the challenge entirely. The persisted source of truth (`identity.authFactor` / `user.authFactor`) is already on the loaded document but was ignored.

## Fix

- **Login enforcement** (`signIdentity`, `signUser`): decide based on the persisted `authFactor` field, and lazily `register()` it into the local map if missing so `start()`/`authenticate()` work on the handling replica. Every replica now self-heals from the database. Identity and user are fixed symmetrically.

- **Deactivation** (`deleteFactor`, identity + user): `deleteIdentityFactor`/`deleteUserFactor` only clear the *in-progress activation* map, which is empty once activation has completed. The old code threw `NotFound` based on that map and **skipped the DB `$unset` entirely**, so deactivation never persisted (it only appeared to work because the in-memory map was cleared first — which doesn't propagate across replicas either). Existence is now derived from the persisted `findOneAndUpdate` result, and the `$unset` always runs.

## Tests

- New regression test: after activating 2FA, the instance's in-memory factor state is cleared (reproducing a replica that missed the broadcast); login must still return a challenge — and the subsequent TOTP answer must succeed.
- Strengthened the deactivation test to clear in-memory state after deactivation, proving the removal is persisted rather than only cleared from one instance's map.
- Full `api/passport` suite: 85 passed, 1 skipped.

### DIFF
--- a/packages/api/passport/identity/src/identity.controller.ts
+++ b/packages/api/passport/identity/src/identity.controller.ts
@@ -219,14 +219,19 @@ export class IdentityController {
     )
   )
   async deleteFactor(@Param("id", OBJECT_ID) id: ObjectId) {
-    const res = this.deleteIdentityFactor(id.toHexString());
+    // deleteIdentityFactor only clears any in-progress activation; an already-activated factor is
+    // no longer in that transient map, so existence must be derived from the persisted record.
+    this.deleteIdentityFactor(id.toHexString());
     this.authFactor.unregister(id.toHexString());
 
-    if (!res) {
+    const identity = await this.identityService.findOneAndUpdate(
+      {_id: id},
+      {$unset: {authFactor: ""}}
+    );
+
+    if (!identity) {
       throw new NotFoundException(`Identity with ID ${id} not found`);
     }
-
-    await this.identityService.findOneAndUpdate({_id: id}, {$unset: {authFactor: ""}});
   }
 
   @Get(":id")
--- a/packages/api/passport/src/passport.identity.controller.ts
+++ b/packages/api/passport/src/passport.identity.controller.ts
@@ -179,7 +179,14 @@ export class PassportIdentityController {
     const refreshTokenSchema = await this.identityService.signRefreshToken(identity);
 
     const id = identity._id.toHexString();
-    if (this.authFactor.hasFactor(id)) {
+    // Enforce 2FA from the persisted record, not the in-memory AuthFactor map: that map is
+    // per-replica and only synced opportunistically, so a replica that missed the registration
+    // broadcast would otherwise skip the challenge and let the user bypass 2FA.
+    if (identity.authFactor) {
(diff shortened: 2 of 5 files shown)


htuna07:
Addressed all three review findings in 88c0fcc3c:

**Issue 1 — SSO completion crash (the important one).** `completeIdentifyWithState` / `completeLoginWithState` now handle the `factorRes` branch and return the challenge, mirroring the credentials `_identify`/`_login` path:

```ts
const {tokenSchema, refreshTokenSchema, factorRes} = await this.signIdentity(identity, expires);
if (factorRes) {
  return res.status(200).json(factorRes);
}
this.setRefreshTokenCookie(res, refreshTokenSchema.token);
res.status(200).json(tokenSchema);
```

I went with returning the challenge rather than a hard 400: the held state-request is the same long-poll `res` used by the credentials flow, and `signIdentity`/`signUser` already populate the pending-token map, so the existing `/factor-authentication` round-trip works for SSO too. This enforces 2FA for SSO logins instead of locking those users out — and fixes the TypeError.

**Issue 2 — user-flow coverage.** Added a parallel regression test for `POST /passport/login` that mirrors the identity test (activate 2FA, clear the in-memory map to simulate a replica that missed the broadcast, assert a challenge is still issued and the TOTP answer succeeds). To make `POST /passport/user` validate in the shared test module (user-create declares a `date-time` field), I registered `OBJECT_ID`/`DATE_TIME` formats in its `SchemaModule.forRoot()`.

**Issue 3 — typo.** Fixed `res.headerSent` → `res.headersSent` in the user SSO completion guard.

Full `api/passport` suite: **86 passed, 1 skipped**.

htuna07:
@claude fix "One residual typo — same bug, different location" you mentioned on your review

idriska (approved):

PR #1875, opened 2026-06-26
fix(panel): remove redundant paginate:true from bucket search query

### LINKED ISSUE #1871: Bucket data list sends pagination=true to backend but UI has no pagination feature
## Bug Description
Bucket data list page sends the pagination parameter as `true` to the backend, but the UI does not have a pagination feature. The pagination parameter should either be omitted from the request or explicitly set to `false`.

## Steps to Reproduce
1. Navigate to the Bucket data list page.
2. Observe the network request sent to the backend.
3. Note that `pagination=true` is included as a query parameter.

## Expected Behavior
The `pagination` parameter should be omitted or set to `false`, as the UI does not support or expose any pagination controls.

## Actual Behavior
`pagination=true` is sent to the backend despite no pagination UI being present.

## Environment
- Project: Spica
- Area: Bucket / Data List Page

htuna07:
The search query built in `useBucketSearch` already inherits `paginate:true` from `useBucketData`, which always sets it before spreading the search query. Having it duplicated in the search query object was misleading and unnecessary.

**Note:** The `paginate: true` in `useBucketData.ts` is intentional — the bucket data list has an infinite scroll (load-more) feature that requires `meta.total` from the paginated backend response to know when all rows are loaded.

Fixes #1871

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/apps/panel/src/hooks/useBucketSearch.ts
+++ b/apps/panel/src/hooks/useBucketSearch.ts
@@ -5,7 +5,6 @@ const escapeForRegex = (s: string) => s.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
 
 const buildBucketQuery = (searchText: string, searchableColumns: string[]) =>
   ({
-    paginate: true,
     relation: true,
     limit: 25,
     filter: {


htuna07:
@claude why "Note: The paginate: true in useBucketData.ts is intentional — the bucket data list has an infinite scroll (load-more) feature that requires meta.total from the paginated backend response to know when all rows are loaded."? Especially why meta.toal is needed to know all rows are laoded? Couldn't be done by filter + skip on each scroll and ends when server returns nothing?

PR #1899, opened 2026-07-14
perf(function): rolling worker cutover on function update

htuna07:
## Problem

When a function's baked runtime inputs change under load (index/code, dependencies, env/secret values, or timeout), the engine immediately flipped every live worker of that function to `Outdated`. Outdated workers are invisible to the scheduler, so the very next event had to **cold-spawn** a `Fresh` worker that lazily imports the new module. Between the update and the first warm graduation, every event paid a cold module-load cost and the function's throughput collapsed.

## Change

Turn the hard cutover into a **rolling** one:

1. **Pre-warm** one fresh replacement worker per currently-serving worker — independent of the per-function `warmWorkers` setting (works even when `warmWorkers: 0`).
2. Mark live workers **superseded** ("before-outdated") instead of outdating them — they keep serving the old code so there's no cold-start gap.
3. On assignment, prefer a ready warm replacement; when one takes over, **retire one superseded worker**. If none is ready yet, keep serving from the still-hot superseded worker rather than cold-spawning.
4. **Safety valve:** a grace timer (`--function-worker-cutover-grace-ms`, default 30s) force-retires surviving superseded workers if replacements never become ready (e.g. a version that crashes on preload), so a broken deploy surfaces its error instead of being masked by stale code.

The old warm reserve is retired **synchronously** on update (flipped to `Outdated` before anything else), and `canServeWarm` gates on `Warm` state — so an old warm worker is never picked after an update; only fresh warm workers (replacements + the refreshed steady-state reserve) are servable. `complete()` now reaps drained retired workers, which also fixes a pre-existing lingering non-warm-outdated worker.

### New worker-selection order (`takeAWorker`)
`canServe` (hot new-code reuse) → `canServeWarm` (reserve **or** replacement) → `canServeSuperseded` (hot old code) → `Fresh` cold fallback.

## Files
- `scheduler/src/worker.ts` — `superseded`/`replacement` flags, `canServeSuperseded`, `canServe` excludes superseded.
- `scheduler/src/scheduler.ts` — `supersedeWorkers`, replacement pool, `retireOneSuperseded`, grace timer, `takeAWorker`/`process`/`complete` updates, `scaleWarmWorkers` excludes replacements, status count.
- `src/plan-executor.ts` — routes `plan.outdate` through the fresh-target reconcile.
- `interface/function/scheduler`, `main.ts`, `function.module.ts` — `functionWorkerCutoverGraceMs` option.

## Design decisions
- **Flag, not a new `WorkerState`** — a superseded worker is still `Targeted`/`Busy` and tracks capacity normally, so a flag keeps the state machine intact and confines changes to routing predicates.
- **Pre-warm count = active (serving) workers.**
- **Grace-timeout-then-force-cutover** for broken deploys (over pure rolling), to avoid masking a bad deploy.

## Tests
- `api/function/scheduler`: **62/62 pass** — 6 worker-state tests + 8 scheduler cutover tests (incl. explicit `warmWorkers > 0` proofs that the old reserve is retired and only fresh warm workers are servable).
- `api/function`: **159/159 pass** — updated the two tests that asserted `outdateWorkers` on active-trigger updates to expect `supersedeWorkers`.
- `yarn nx build api`: compiles clean.

### DIFF
--- a/packages/api/function/scheduler/src/scheduler.ts
+++ b/packages/api/function/scheduler/src/scheduler.ts
@@ -574,19 +574,22 @@ export class Scheduler implements OnModuleInit, OnModuleDestroy {
    * Roll a function's workers over to fresh state without a cold-start gap. Instead of outdating
    * live workers immediately (which forces the next event to cold-spawn), it keeps them serving the
    * old code as "superseded" and pre-warms one fresh replacement per serving worker. `takeAWorker`
-   * then prefers the replacements; as each is consumed, one superseded worker is retired. The old
-   * warm reserve is killed here (it can't drain by serving) and refilled from fresh state by
-   * reconcileWarmWorkers afterwards.
+   * then prefers the replacements; as each is consumed, one superseded worker is retired. Every
+   * pre-warmed worker (the steady-state reserve and any in-flight replacements from an earlier
+   * update) is killed here since it holds now-stale code; the reserve is refilled by
+   * reconcileWarmWorkers afterwards and replacements are respawned from this newest target.
    */
   supersedeWorkers(target: event.Target) {
     const fnId = target.id;
     const workers = Array.from(this.workers.values()).filter(worker => worker.hasSameTarget(fnId));
 
+    // Kill every pre-warmed worker of this function — the steady-state reserve AND any in-flight
+    // replacements from an earlier cutover. All of them preloaded now-stale code (a second update
+    // landing mid-warming makes the first update's replacements stale), so none can be reused:
+    // reconcileWarmWorkers refills the reserve and scaleReplacementWorkers respawns replacements,
+    // both from the newest target.
     for (const worker of workers) {
-      if (
-        !worker.isReplacement &&
-        (worker.state == WorkerState.Warm || worker.state == WorkerState.Warming)
-      ) {
+      if (worker.state == WorkerState.Warm || worker.state == WorkerState.Warming) {
         worker.markAsOutdated();
         worker.kill();
       }
--- a/packages/api/function/scheduler/src/scheduler.ts
+++ b/packages/api/function/scheduler/src/scheduler.ts
@@ -568,6 +568,11 @@ export class Scheduler implements OnModuleInit, OnModuleDestroy {
           }
         }
       });
+
+    // A hard outdate ends any rolling cutover in progress for this function (it's being deleted or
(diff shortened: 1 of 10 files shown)


packages/api/function/scheduler/src/scheduler.ts
```diff
@@ -510,6 +568,185 @@ export class Scheduler implements OnModuleInit, OnModuleDestroy {
...
+      .filter(
+        worker =>
+          worker.hasSameTarget(fnId) &&
+          worker.isReplacement &&
+          (worker.state == WorkerState.Warm || worker.state == WorkerState.Warming)
+      )
+      .forEach(worker => {
+        worker.markAsOutdated();
+        worker.kill();
+      });
```
htuna07, line 697: Why are these lines in this method? Are they even necessary?

packages/api/function/src/plan-executor.ts
```diff
@@ -123,8 +131,20 @@ export class PlanExecutor {
...
+    // nothing serving, so fall back to a plain outdate.
+    if (toSupersede.has(functionId)) {
+      if (hasActiveTrigger) {
+        this.scheduler.supersedeWorkers(target);
+      } else {
+        this.scheduler.outdateWorkers(functionId);
+      }
+    }
+
+    this.scheduler.reconcileWarmWorkers(target, hasActiveTrigger ? fn.warmWorkers ?? 0 : 0);
```
htuna07, line 147: These changes make outdating and reconciliation plan executions coupled, can't be handled better?

packages/api/function/scheduler/src/scheduler.ts
```diff
@@ -425,7 +462,15 @@ export class Scheduler implements OnModuleInit, OnModuleDestroy {
...
-      this.workers.get(workerId)?.onComplete();
+      const worker = this.workers.get(workerId);
+      worker?.onComplete();
+
+      // A retired (superseded → outdated) worker that just drained its last in-flight event is
+      // done and would otherwise linger; reap it now. Guarded to Outdated so live workers, which
+      // go Busy→Targeted here and get reused, are never killed.
+      if (worker && worker.state == WorkerState.Outdated && worker.isIdle()) {
+        worker.kill();
+      }
```
htuna07, line 473: why such a change? won't be killed automatically when the timeout comes?

idriska (approved):

PR #1908, opened 2026-07-20
perf(preference): replace preferences change stream with in-memory dispatcher

htuna07:
Follow-up to #1907, applying the same treatment to **preferences**.

## Context

`PreferenceService.watchPreference(scope)` opened a MongoDB change stream on the `preferences` collection so consumers could re-read a preference when it changed. That stream is redundant: **every production write to `preferences` goes through `PreferenceService`** — verified by grepping the raw collection name repo-wide. In fact there is a single production write site, `PreferenceController.replaceOne` (`PUT /preference/:scope`).

Replaced with the in-process pub/sub (`CollectionChangeDispatcher`) already used by `env_var`, `secret`, `bucket`, and `webhook`.

**Extra win over the webhook case:** `watchPreference` returned a *cold* observable, so every subscriber opened its own change stream. They now share a single in-memory `Subject`.

Multi-replica correctness is preserved via the optional `ClassCommander` SYNC broadcast.

## Changes

- **New `PreferenceChangeDispatcher`** (`services/src/change-dispatcher.ts`), provided + exported by `PreferenceModule.forRoot()` (already `@Global`, provides the service once).
- **`PreferenceService`** — dispatches on `insertOne` and `replace`. `replace` now defaults `returnDocument` to `AFTER` so the replaced document (and its `_id`) is available to dispatch from; an explicit caller can still override it. All three existing callers are unaffected — the controller already passes `AFTER`, and the two tests ignore the return value.
- **`watchPreference`** rebuilt on the dispatcher. Consumers key off `scope` but the event carries only `_id`, so it reloads by `_id` and filters on `scope` — exactly what the old `$match: {"fullDocument.scope": ...}` did. **Deletions still do not propagate**, matching the previous behaviour (delete events carried no `fullDocument`).
- Reloads are **serialized on a promise chain**, so the `propagateOnStart` value cannot arrive after a change and two rapid updates cannot emit out of order — which would otherwise leave `BucketSchemaResolver` compiling against a stale preference.

## Behaviour left untouched

- `default()` is in-memory only and never wrote to Mongo — unchanged.
- The `0.16.0` migration writes `preferences` directly, but migrations run as a separate offline process with no API watcher live.
- `PreferenceTestingModule` supplies a fake service via `useValue` and never constructs the real one, so the ~45 specs depending on it are unaffected.

## Tests

The only real consumer chain — `BucketService.watchPreferences()` → `BucketSchemaResolver` — had **no real coverage** (its own spec stubs `watchPreferences`). Added a regression test asserting the bucket schema recompiles when a language is added: posting a document in a not-yet-available language is rejected, and accepted after `PUT /preference/bucket` adds it.

I verified the test has teeth by temporarily disabling the dispatch — it fails (400 instead of 201). Notably the pre-existing "language deleted" test still passed under that break, confirming it only covered the finalizer, not the watch.

The service spec's propagation test moved off the inherited `updateOne` (which no longer dispatches) to `replace`, dropping a `setTimeout`-based race.

## Verification

- `yarn nx test api/preference` → **12 passed, 4 suites**
- `yarn nx run-many -t test -p api/bucket/test api/bucket/services` → **150 passed, 17 suites**
- `yarn build:api` → compiles clean

### DIFF
--- a/packages/api/preference/services/index.ts
+++ b/packages/api/preference/services/index.ts
@@ -1,1 +1,2 @@
 export {PreferenceService} from "./src/preference.service.js";
+export {PreferenceChangeDispatcher} from "./src/change-dispatcher.js";
--- a/packages/api/preference/services/src/change-dispatcher.ts
+++ b/packages/api/preference/services/src/change-dispatcher.ts
@@ -1,0 +1,9 @@
+import {Injectable, Optional} from "@nestjs/common";
+import {ClassCommander, CollectionChangeDispatcher} from "@spica-server/replication";
+
+@Injectable()
+export class PreferenceChangeDispatcher extends CollectionChangeDispatcher {
+  constructor(@Optional() commander?: ClassCommander) {
+    super(commander);
+  }
+}
--- a/packages/api/preference/services/src/preference.service.ts
+++ b/packages/api/preference/services/src/preference.service.ts
@@ -1,4 +1,4 @@
-import {Injectable} from "@nestjs/common";
+import {Injectable, Logger} from "@nestjs/common";
 import {
   BaseCollection,
   Collection,
@@ -17,6 +17,7 @@ import {PreferenceChangeDispatcher} from "./change-dispatcher.js";
 
 @Injectable()
 export class PreferenceService extends BaseCollection("preferences") {
+  private readonly logger = new Logger(PreferenceService.name);
   private _defaults = new Map<string, Preference>();
 
   constructor(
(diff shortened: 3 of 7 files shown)


htuna07:
Addressed the review finding in c857f1d9.

**The bug was real.** `chain.then(onFulfilled)` with no rejection handler means one rejected reload leaves `chain` permanently rejected, and every later link is skipped — so the subscription silently stops propagating. It doesn't self-heal, because `BucketSchemaResolver` subscribes once in its constructor and shares that subscription for the process lifetime via `shareReplay(1)`.

Fixed slightly differently than suggested: rather than swallowing errors with an empty `catch`, each link absorbs its failure **and logs it** via `Logger`, matching the existing idiom in `BucketService` (which logs errors from its own change-dispatcher watch). Silently dropping the update would leave the resolver serving a stale preference with no signal anywhere. Same guard applied to the `propagateOnStart` seed.

Added a regression test (`should keep propagating after a reload fails`) that forces the first `findOne` to reject and asserts a later change still propagates. Verified it has teeth — with the `try/catch` removed it hangs until the 30s timeout, which is exactly the production symptom.

On the minor point: the PR description no longer mentions a "fallback lookup" — that was removed in an earlier push, so the review was reading a stale description.

**Note on scope:** the identical bug is in `WebhookService.targets()`, which is already merged to master via #1907. Fixing that separately so this PR stays single-purpose.

`yarn nx test api/preference` → 13 passed, 4 suites. `yarn build:api` clean.

idriska (approved):

PR #1916, opened 2026-07-29
fix(charts): make replica set bootstrap and credentials robust

htuna07:
## Problem

A Spica install can come up with a healthy MongoDB pod that no client can reach. The API and the migrate job both fail with:

```
MongoServerSelectionError: getaddrinfo ENOTFOUND <release>-database-0
  reason: TopologyDescription { type: 'ReplicaSetNoPrimary', setName: 'rs0',
    servers: Map(1) { '<release>-database-0:27017' => [ServerDescription] } }
```

Two independent defects combine to produce it.

**1. The replica set is bootstrapped with an unresolvable host.** `init-users.sh` called an argument-less `rs.initiate()`, so MongoDB derived the member host from `hostname` — a bare pod name that resolves only inside the database pod itself. Clients connect to the correct FQDN seed, run `hello`, are handed `<release>-database-0:27017` by SDAM, discard the good seed for it, and fail. `mongod` itself stays healthy and `PRIMARY` throughout, which makes this look like a networking problem rather than a config one.

Recovering depended entirely on the replication controller landing a reconfig afterwards.

**2. That recovery path breaks on roughly 1 install in 60.** `generatePassword` shuffles a charset containing `-`. When `-` lands first, the password is parsed as a *flag* rather than a value by both mongosh and yargs:

```
Error parsing command line: unrecognized option: -kYpQ0!8TTsC
```

Every `--password` consumer in the chart is affected: the replication controller, the startup probe, and `set-profiling-level.sh`. In the controller this error matches neither `connection attempt failed` nor `ENOTFOUND`, so it re-throws immediately and burns `backoffLimit: 25` without ever reconfiguring — leaving defect 1 permanently uncorrected.

Observed across 7 namespaces on one cluster: the single namespace whose root password began with `-` was the single namespace whose controller job failed and whose replica set was still on the bare hostname.

## Changes

- **`initdb.yaml`** — initiate with an explicit FQDN member from a new `RS_PRIMARY_HOST` env. The config is now correct at creation and no longer depends on the controller. Correct for single-node, and the right seed for multi-node where the controller still adds secondaries.
- **`_helpers.tpl`** — new `database.primaryNode` helper; `generatePassword` keeps its first character alphabetic.
- **Credentials passed as `--flag=value`** in the controller, the startup probe, the controller job args, and `set-profiling-level.sh`. This is what actually makes the chart immune, since existing installs keep their password via `lookup`.
- **`set-profiling-level.sh`** additionally used `--eval` with no argument followed by a heredoc, so profiling was never applied.
- **`main.ts`** — besides the credential form, all calls go through `mongoAddress()` with an explicit `directConnection=true`. This one is **behaviour-neutral today** and is not fixing a live bug: mongosh already resolves a lone `--host` to a `Single` (direct) topology, verified against a live instance. It pins that implicit default, which has changed across driver majors, for a tool whose whole job is addressing one specific node. Happy to drop it if you'd rather keep the PR minimal.

## Verification

- 159/159 helm-unittest tests pass; `helm lint`, `tsc`, and Prettier clean.
- New regression tests: `init-users.sh` uses the FQDN and never an argument-less initiate; the init container receives `RS_PRIMARY_HOST`.
- Password generator: 60/60 renders produce an alphabetic first character.
- Both command shapes exercised against a live affected database, using the actual `-`-leading password that caused the outage:

```
OLD:  Error parsing command line: unrecognized option: -kYpQ0!8TTsC
NEW:  {"status":1,"primary":true,"host":"...svc.cluster.local:27017"}
```

## Note

`rs0` is hardcoded in `mongod --replSet` while `.Values.database.replicaSetName` exists and is consumed only by the application, so setting that value to anything else is already broken. This PR matches the hardcode rather than widen scope — worth a separate fix.

### DIFF
--- a/apps/mongoreplicationcontroller/src/main.ts
+++ b/apps/mongoreplicationcontroller/src/main.ts
@@ -84,15 +84,17 @@ function authenticateForMongo(username, password, authenticationDatabase) {
   debug(
     "Assuming mongodb started with --auth flag and credentals will be provided for each connection."
   );
-  mongoCommand = `${mongoCommand} --username '${username}' --password '${password}' --authenticationDatabase '${authenticationDatabase}'`;
+  mongoCommand = `${mongoCommand} --username='${username}' --password='${password}' --authenticationDatabase='${authenticationDatabase}'`;
+}
+
+function mongoAddress(host: string) {
+  return `"mongodb://${host}/admin?directConnection=true"`;
 }
 
 async function findPrimaryNode(nodes: string[]) {
   for (const node of nodes) {
     try {
-      const {stdout} = await execMongo(
-        `admin --host "${node}" --eval "JSON.stringify(db.hello())"`
-      );
+      const {stdout} = await execMongo(`${mongoAddress(node)} --eval "JSON.stringify(db.hello())"`);
       debug(stdout);
       const result = JSON.parse(stdout);
       if (result.isWritablePrimary === true) {
@@ -113,7 +115,9 @@ async function findPrimaryNode(nodes: string[]) {
 async function initiateReplication(nodes: string[], reinitiate = false) {
   let statusResult;
   try {
-    const {stdout} = await execMongo(`--host "${nodes[0]}" --eval 'JSON.stringify(rs.status())'`);
+    const {stdout} = await execMongo(
+      `${mongoAddress(nodes[0])} --eval 'JSON.stringify(rs.status())'`
+    );
     statusResult = JSON.parse(stdout);
   } catch (error) {
     debug(error);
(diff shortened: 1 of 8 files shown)


htuna07:
Superseded by #1917, which is narrowed to the one defect that actually caused the outage — the password generator emitting a leading `-`.

Closing this because most of what it contained does not hold up:

- **`findPrimaryNode` reporting a secondary as primary** — false. `mongosh --host <single host>` already resolves to a `Single` (direct) topology, verified against a live instance, so the `directConnection` change was behaviour-neutral.
- **`--eval` with a heredoc breaking profiling** — false. mongosh falls through to reading stdin as a script; it exits 0 and `getProfilingStatus()` returns the intended `{was: 1, slowms: 100}`.
- **The startup probe being affected by the password** — false. Kubernetes does not expand `$(VAR)` in probe exec commands, so the probe runs a literal `$(ROOT_PASSWORD)` that can never start with `-`. It passes only because `ping` is permitted pre-auth.

Those three changes fixed nothing, and two of them forced a MongoDB rolling restart on every existing installation by altering the StatefulSet pod template and its `checksum/initdb` annotation.

The remaining item — bootstrapping the replica set with an explicit FQDN instead of a bare hostname — is real, but it is hardening rather than a bug fix: the replication controller does normally repair the bare config. It also forces a restart, so it deserves its own PR on a schedule where database downtime is expected, together with the restart-free `--flag=value` credential changes.

#1917 is a single-file change to the password generator and is verified to leave installed releases byte-identical.

PR #1942, opened 2026-09-09
fix(passport): point strategy callback urls at the scoped routes

BatuhanEvirgen:
OAuth and SAML strategies still advertised the pre-split passport/strategy/:id/complete route, which this build no longer serves: oauth is handled by passport/user/... and saml by passport/identity/... Every OAuth login broke at the callback with a 404, and the value could not be corrected through the API because afterInsert rewrites it on every update.

The e2e tests missed this because they threw away the path of the advertised redirect_uri and hardcoded the endpoint instead; they now follow the url the strategy itself hands out.

Existing strategies keep the stale redirect_uri in the database, so a migration rewrites them.

Claude-Session: https://claude.ai/code/session_016FHYUQchKyt5bieGHZEGLq


htuna07:
@claude why did the claude code review on this pr stuck for a long time, fix the issue

htuna07 (approved):

PR #1943, opened 2026-09-09
fix(panel): restore bucket table infinite scroll

BatuhanEvirgen:


BatuhanEvirgen:
@claude review has been running for almost an hour without completing. Could you please check and fix that?

BatuhanEvirgen:
@claude The previous review was based on an older commit. I’ve fixed the double-fire race issue, so could you please run a new review against the latest changes?

htuna07 (approved):

PR #1946, opened 2026-09-18
fix(bucket): clear removed fields of buckets owning a property named type

BatuhanEvirgen:
Converting a schema path into a document path guessed whether a segment was a schema node by checking for a "type" key on it. A bucket property named "type" puts that key on the properties map itself, so the map was mistaken for a property definition and "properties" leaked into the document path.

Every field removal and type change in such a bucket then unset "properties.<field>" instead of "<field>", which matches nothing, so the values stayed in the collection. Patching those documents afterwards failed validation with "must NOT have additional properties".

Walk the schema structurally instead: consume "properties" together with the field name that follows it, and "items" on an array node. Both the document path and the trailing keyword path now come from the same walk, so a field can no longer be confused with a keyword of the same name.


BatuhanEvirgen:
@claude The review has been running for over 20 minutes without completing. Please investigate what's causing the delay, fix the issue, and complete the review.

BatuhanEvirgen:
@claude  The review appears to be stuck again. Fix that issue and complete the review.

htuna07 (approved):

